Issue #11, opened 2016-07-29
Task 3.2: Unit tests and CI tests for recursive support for performant upload/download


rrenfrow86:

begoldsm:
Closing as this is done.

Issue #9, opened 2016-07-29
Task 2.2: Unit tests and CI tests for multi-part upload/download for single files


rrenfrow86:

begoldsm:
Closing as this is done, thanks!

Issue #6, opened 2016-07-27
Task 1.2: Unit tests and CI tests for core DLS filesystem client


rrenfrow86:

begoldsm:
Closing this issue since it is resolved and merged.

PR #40, opened 2016-08-21
PR: Record test requests/responses

jbcrail:
These are the features included to get a green Travis build:

* reduced fixture size
* fixed spurious newlines for some CLI commands
* isolated tests to allow parallel builds for each Python version
* recorded HTTP requests/responses
* cached build dependencies for faster test startup
* fixed several broken tests due to parallel builds
* refactored to use singleton for default test working directory (in lieu of hard-coded value)
* removed need for valid Azure account information

The following were excluded and can be covered in future PRs:

* add coverage support
* fill in missing tests needed for #6, #9, and #11


martindurant:
Green "PASSED" labels - good job!

tests/testing.py
```diff
@@ -54,7 +54,7 @@ def _scrub_sensitive_response_info(response):
 
 my_vcr = vcr.VCR(
     cassette_library_dir=recording_path,
-    record_mode="once",
+    record_mode=settings.RECORD_MODE,
```
martindurant: Can we have some documentation (maybe in a file, not in the actual documentation) of how to add new tests with recorded server responses?

martindurant:
Recommend merge.

A command like `py.test --cov=adlfs tests/ -vv` can already do coverage (if pytest-cov is installed), I get 83% on this branch. (you could mark all those help_* functions in CLI as `# pragma: no cover`

tests/README.md
```diff
@@ -0,0 +1,32 @@
...
+To record responses for all tests even if previous recordings exist:
+
+    RECORD_MODE=all py.test -x -vvv --doctest-modules --pyargs adlfs tests
+
+When recording new responses, you will need valid Azure credentials. The following
+environment variables should be defined:
+
+* `azure_username`
+* `azure_password`
+* `azure_store_name`
```
begoldsm: Minor: this should be updated to reflect that it is an ADLS (Azure Data Lake Store) account name, since there could be confusion with the actual "Azure Store" which is not the same thing.

tests/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+If you test the management API, you will need to define:
+
+* `azure_subscription_id`
+* `azure_resource_group_name`
+
+Optionally, you may need to define `azure_tenant_id` or `azure_url_suffix`.
```
begoldsm: Should be an adls url suffix, or perhaps just have some more help text defining what these are used for?

PR #41, opened 2016-08-28
PR: Support Windows filesystem paths

jbcrail:
This adds support for system-independent paths.

The library will now use the system's path representation and convert to POSIX-style paths when talking to Azure.

In addition to Travis, AppVeyor is supported for Windows builds.

Depends on #40


.travis.yml
```diff
@@ -13,6 +13,7 @@ matrix:
 
 install:
   # Install dependencies
+  - pip install --upgrade setuptools
```
martindurant: Why was this required?

jbcrail: By default, Travis's Ubuntu VM uses setuptools 12.0, which doesn't support conditional dependencies without jumping through hoops (see [Conditional Python Dependencies](https://hynek.me/articles/conditional-python-dependencies/)). To keep things simple, I upgraded setuptools to conditionally install the backported `pathlib` package for 2.7 and 3.3. It's included in 3.4+.

Since this is my first time creating a public PyPI package, any suggestions on best practices would be appreciated.

martindurant:
The whole of the previous PR shows up here too, making the diff hard to read :(  (lets get PR #40 merged)

Can you give a brief rationale for swapping out strings for pathlib objects? I'm not convinced that it makes the code any simpler, and of course it does complicate the install requirements. This should make using windows paths easier? I didn't spot tests against paths in windows format. 
Should some documentation be written that people using the code would need?

I've not had to create conditional instals like that, only on the conda side (where it's easy).

jbcrail:
On Windows, the benefits are:
* paths are case-insensitive
* `adlfs.core.glob` is simpler and avoids double-escaping backslash issues when creating glob pattern
* remote paths can be defined using either backslashes or forward slashes (Azure handles this already but now the library also does)
* if we switch to pathlib for local paths, we can eliminate `glob` and `os.path` imports and have consistent functionality, including support for `**` on all versions.

On all systems, the benefits are:
* users can pass both strings and Path/PurePath objects (similar to PEP 519, 'Path-like' objects)
* syntactic sugar for path concatenation using division operator

martindurant:
OK, sounds good.
Had better put at least a line in docs saying that windows and posix paths are handled seamlessly, perhaps give an explicit window path example somewhere.

martindurant:
So in using the `AzureDLFileSystem`, we can use either a posix-like path string, or a path object (which could have been instantiated with either windows- or posix-like paths) - correct? For the CLI, this is done automatically for either kind of path?

adlfs/core.py
```diff
@@ -702,3 +706,50 @@ def _fetch_range(rest, path, start, end, max_attempts=10):
...
+        return s in self.as_posix()
+
+    @property
+    def globless_prefix(self):
+        """ Return shortest path prefix without glob quantifiers. """
+        def no_quantifier(s):
+            quantifiers = ['*', '?']
+            return all(q not in s for q in quantifiers)
+
+        return pathlib.PurePath(*itertools.takewhile(no_quantifier, self.parts))
```
martindurant: I'm sure this can be done without the inner function ;)

martindurant:
```
prefix = []
for piece in self.parts:
    if any(q in piece for q in ['*', '?']):
        break
    prefix.append(piece)
```

jbcrail:
Yes, `AzureDLFileSystem` supports both path strings and path objects. You can also pass a Windows-style path string when on Windows (no path object needed). The CLI doesn't create a path object and only handles path strings.

martindurant:
OK, +1

docs/source/index.rst
```diff
@@ -80,6 +80,32 @@ python files. The recommended way to use this is with a context manager
     with adl.open('newfile', 'rb') as f:
         df = pd.read_csv(f) # read into pandas.
 
+To seamlessly handle remote path representations across all supported platforms,
+the main API will take in numerous path types: string, Path/PurePath, and
+AzureDLPath. On Windows in particular, you can pass in paths separated by either
+forward slashes or backslashes.
```
begoldsm: For clarity, these are just the paths for the remote files, or also for local files for upload? For example, can a windows user put in the local file path /my/file/for/upload.txt? How would that be properly translated to a file on the local NTFS filesystem?

jbcrail: The paths are just for remote files. For local paths, we keep the native path style, so `/my/file/for/upload.txt` would be a file in the current directory with forward slashes in the name.

PR #44, opened 2016-08-31
PR: Memory safe upload

### LINKED ISSUE #42: Uploaded temp files should be 256mb in size
The chunking process needs to be two tiered:
1. At most 4mb of data is uploaded in a single http request
2. The intermediate files should be as close to 256mb as possible.

This is important for a couple of reasons:
1. 4mb upload chunk size is the only way to guarantee atomicity of the upload operation for a single request. For example, if you upload 16mb in a single request and it fails, on retry you need to see if 0, 4, 8 or 12mb of data were uploaded (since the server will always commit in 4mb chunks).
2. The intermediate files should be as closed to 256mb as possible to avoid massive fragmentation and lifecycle performance problems once the data is in the system. 256mb is the optimal size in our system for chunks to be concatenated at.

So, for example, if you have a 10GB file it should be split into ~40 chunks of 256mb. Each of those chunks can be uploaded in parallel using <= 4mb requests.

### LINKED ISSUE #43: Lots of memory errors when attempting to upload/download large files
When attempting to upload or download a large file (such as a 10GB file) I am constantly running into MemoryErrors. Looking at how much memory Python is using during this time I see that it is right around 2GB. This seems like a lot of memory taken to process a 10GB file. Additionally, 2GB of memory use shouldn't be enough to cause these failures in the python runtime (although I imagine that is just a setting somewhere). Ideally we shouldn't be scaling memory usage to be 20% of the file size, since we need to support the ability to upload files in the terabyte range (for upload and download).

martindurant:
Fixes #43 , Fixes #42 

@jbcrail , please see if you think this approach should work. If yes, we can apply it to downloads too (although, no need for small data transfer on each request, which only adds overhead).


martindurant:
@begoldsm , this might solve your memory problem, and upload in 4MB chunks.

martindurant:
@jbcrail , any idea why py33 is failing, yet everything else passes?

begoldsm:
@martindurant thanks for the quick turnaround! Can you resolve the conflicts from my merging of #41 ?

martindurant:
@begoldsm , synced with dev.

PR #63, opened 2016-09-09
PR: Rest updates

### LINKED ISSUE #60: Optimization for APPEND, CREATE and OPEN
To avoid redirects and extra web calls, the following query parameters should be added to the APPEND, CREATE and OPEN operations:
APPEND: append=true
CREATE: write=true
OPEN: read=true

### LINKED ISSUE #61: Replace CONCAT with MSCONCAT
There is a design limitation in concat due to passing all of the paths in on the URI which can cause failure when concatenating a large number of files. Until this is fixed in a new WEBHDFS protocol, please use the MSCONCAT op code and pass in the paths in the body of the request as an application/octet-stream using the following format for the stream in the body:
sources=<path1>,<path2>,<path3>...,<pathN>

To see more information about this API you can see the swagger specification for it here:
https://github.com/Azure/azure-rest-api-specs/blob/master/arm-datalake-store/filesystem/2015-10-01-preview/swagger/filesystem.json#L335

Additionally, for further optimizations for concat please see issue: #49

martindurant:
Fixes #60 ,
Fixes #61


martindurant:
I'll do some performance tests on this yet.

begoldsm:
@martindurant sounds good. I look forward to the results!

adlfs/core.py
```diff
@@ -314,8 +314,13 @@ def mv(self, path1, path2):
...
         outfile = AzureDLPath(outfile).trim()
-        self.azure.call('CONCAT', outfile.as_posix(),
-                        sources=','.join(filelist))
+        filelist = ','.join(AzureDLPath(f).as_posix() for f in filelist)
+        #self.azure.call('CONCAT', outfile.as_posix(),
+        #                sources=','.join(filelist))
+        #self.invalidate_cache(outfile)
+        #return
+        self.azure.call('MSCONCAT', outfile.as_posix(),
+                        data='sources='+filelist, deleteSourceDirectory='true')
```
begoldsm: I would be careful setting this to always be true. The caller should be able to decide, since there could potentially be other files in the source directory they don't want deleted, and this will delete them. This optimization should really only be used by our uploader tools, and only if we can reasonably guarantee that the only files in the folder belong to the one target file being created. This goes back to my recommendation of creating unique folders specific to the file being uploaded, instead of using a single "tmp" folder for all uploads.

martindurant: Ah, now I understand what you mean.

adlfs/core.py
```diff
@@ -311,11 +311,27 @@ def mv(self, path1, path2):
...
+        """ Concatenate a list of files into one new file
+
+        Parameters
+        ----------
+
+        outfile : path
+            The file which will be concatenated to. If it already exists,
+            the extra pieces will be appended.
+        filelist : list of paths
+            Existing aslfs files to concatenate, in order
```
begoldsm: >aslfs [](start = 21, length = 5)

Minor: adl, not asl

adlfs/multithread.py
```diff
@@ -327,7 +327,7 @@ def run(self, nthreads=None, monitor=True):
         for (rfile, lfile), dic in self.progress.items():
             unique = dic['uuid']
             if len(dic['waiting']) > 1:
-                parts = [self.temp_upload_path+unique+"_%i" % i for i
+                parts = [self.temp_upload_path+unique+"/part_%i" % i for i
```
begoldsm: Does this create a folder for each part that is uploaded? I would think we want to create one folder for all the parts to go into instead, right?

adlfs/core.py
martindurant: fixed.

adlfs/multithread.py
martindurant: It created one folder (identifier `unique`, above) for each multi-chunk file, into which all the parts go. Interestingly, it seems there is no need to explicitly create the folder first.

martindurant:
NB: I will need to re-record the VCR data to get tests to pass.

adlfs/multithread.py
begoldsm: Correct, the first file will result in the folder being created. We have no hard requirement on the underlying directory existing prior to file creation.

---
In reply to: [78467170](https://github.com/Azure/azure-data-lake-store-python/pull/63#discussion_r78467170) [](ancestors = 78467170)

begoldsm:
@martindurant I am signed off on this pending a clean CI run.

Issue #39, opened 2016-08-18
Persisting of up/downloader states


martindurant:
Should decide on cadence and method of saving object state, not just on ^C, and CLI to list/restart objects.

Part of #13

rrenfrow86:
The uploader/downloader is saved on every status change.

begoldsm:
This is correct, thanks!

PR #64, opened 2016-09-13
PR: Downloader/uploader improvements

jbcrail:
This PR will give us improved control of transfer activity and will allow us to save/resume/monitor with more detail. The common transfer functionality has been refactored into a separate module.

Whether uploading or downloading, files can be written into temporary files and merged at the end. Previously, this was only available when uploading.

The actual callables used to transfer and merge can be overridden, making it easier to add unit tests. For example, the transfer/merge callables can be no-ops, thus, testing core functionality without network traffic.
 
Other miscellaneous improvements:
* Add fine-grained states for file and chunk statuses
* Add hooks to view transfer progress (helps with #39)
* Add parameters to tune temporary files
* Add timeout to transfer monitor
* Shutdown transfer more cleanly

~~Note: persistence is currently broken and will be fixed soon.~~


adlfs/transfer.py
```diff
@@ -0,0 +1,312 @@
...
+
+logger = logging.getLogger(__name__)
+
+File = namedtuple('File', 'src dst state nbytes start stop chunks')
+Chunk = namedtuple('Chunk', 'name state offset retries')
+
+
+class DisjointState(object):
+    """
+    Collection of disjoint sets containing state about objects.
```
martindurant: This description doesn't tell me what this class is for. I suspect it is meant to be purely internal.

adlfs/transfer.py
```diff
@@ -0,0 +1,312 @@
...
+        name: str
+            Unique ID used for persistence.
+        transfer: callable
+            Function or callable object invoked when transferring chunks.
+        merge: callable [None]
+            Function or callable object invoked when merging chunks. If None,
+            then merging is skipped.
+        nthreads: int [None]
+            Number of threads to use. If None, uses the number of cores.
+        chunksize: int [2**26]
```
martindurant: This is 32MB. Upload chunks should be 256MB, and download chunks - I'm not sure.

adlfs/transfer.py
```diff
@@ -0,0 +1,312 @@
...
+                 tmp_unique=True, persist_path=None, delimiter=None):
+        """
+        Parameters
+        ----------
+        adlfs: ADL filesystem instance
+        name: str
+            Unique ID used for persistence.
+        transfer: callable
+            Function or callable object invoked when transferring chunks.
+        merge: callable [None]
```
martindurant: Perhaps should state the required call signature here.

adlfs/transfer.py
```diff
@@ -0,0 +1,312 @@
...
+
+class ADLTransferClient(object):
+    def __init__(self, adlfs, name, transfer, merge=None, nthreads=None,
+                 chunksize=2**26, tmp_path='/tmp', tmp_prefix='part_',
+                 tmp_unique=True, persist_path=None, delimiter=None):
+        """
+        Parameters
+        ----------
+        adlfs: ADL filesystem instance
+        name: str
```
martindurant: Can this be auto-generated? (or does that happen in the up/downloader subclasses?)

adlfs/transfer.py
```diff
@@ -0,0 +1,312 @@
...
+            Unique ID used for persistence.
+        transfer: callable
+            Function or callable object invoked when transferring chunks.
+        merge: callable [None]
+            Function or callable object invoked when merging chunks. If None,
+            then merging is skipped.
+        nthreads: int [None]
```
martindurant: We talked about a possible upper maximum for this. I would include a module-level constant (which can still be None) that the user could in theory change.

adlfs/transfer.py
```diff
@@ -0,0 +1,312 @@
...
+            logger.debug("Submitted %s, byte offset %d", name, offset)
+            dic['cstates'][name] = 'running'
+            dic['chunks'][name] = dict(
+                future=self._submit(transfer, self._adlfs, src, name, offset,
+                                    self._chunksize),
+                retries=self._chunkretries,
+                offset=offset)
+
+    @property
+    def progress(self):
```
martindurant: To the discussion, this is where you could have a progress_summary property, which gives total time to far, number of chunks done/doing/to go etc. That's what a progressbar would call. Probably out of scope for this PR

adlfs/transfer.py
```diff
@@ -0,0 +1,312 @@
...
+                    name=name,
+                    state=self._files[key]['cstates'][name],
+                    offset=self._files[key]['chunks'][name]['offset'],
+                    retries=self._files[key]['chunks'][name]['retries']))
+            files.append(File(
+                src=src,
+                dst=dst,
+                state=self._fstates[key],
+                nbytes=self._files[key]['nbytes'],
+                start=self._files[key]['start'],
```
martindurant: This will be the start time of the whole process, no? Should maybe be `time.time()`, since there will always be some lag even if it's one of the chunks read to run immediately. Or perhaps update this when the chunk actually starts its running state. (not very important, just a thought)

adlfs/transfer.py
jbcrail: This is auto-created in the up/downloader classes. Each class defines the client as appropriate (chunk size, merge, persistence path, etc).

The client can also be created outside the up/downloader and passed in. This is especially helpful for testing.

adlfs/transfer.py
```diff
@@ -0,0 +1,312 @@
...
+                    dic['stop'] = time.time()
+                    self._fstates[(src, dst)] = 'finished'
+                    self._status(src, dst, dic['nbytes'], dic['start'], dic['stop'])
+        self.save()
+
+    def run(self, nthreads=None, monitor=True, before_scatter=None):
+        self._nthreads = nthreads or self._nthreads
+        for src, dst in self._files:
+            self._files[(src, dst)]['start'] = time.time()
+            self._fstates[(src, dst)] = 'transferring'
```
martindurant: Should there be a pending state before at least one chunk of a given file has started?

adlfs/transfer.py
```diff
@@ -0,0 +1,312 @@
...
+            if before_scatter:
+                before_scatter(self._adlfs, src, dst)
+            self._scatter(src, dst, self._transfer)
+        if monitor:
+            self.monitor()
+
+    def _cancel(self):
+        for dic in self._files.values():
+            for transfer in dic['chunks'].values():
+                transfer['future'].cancel()
```
martindurant: This does not kill running threads as far as I know - so network activity will continue potentially for some time.

adlfs/transfer.py
jbcrail: Is 64 threads a reasonable maximum? From my experiments, I used up to 64 threads before diminishing returns kicked in.

adlfs/transfer.py
jbcrail: Good call. I'm adding a separate section detailing both signatures.

adlfs/transfer.py
martindurant: I think so - and needs to be mentioned in the docstring here what that constant is

martindurant:
This may not be the right place to mention it (should be a separate small PR), to make upload startup faster, os.walk and stat should be replaced with scandir. I'm commenting here as a reminder - I may have also said this on the relevant issue.

adlfs/transfer.py
jbcrail: The start time is set for each file when the chunks are created and submitted to the thread pool, so it roughly approximates the start time of the whole process. The end time is set when all chunks for a particular file are merged (if applicable) or downloaded. This is a rough measurement, but better bookkeeping is definitely needed if we need more accuracy.

martindurant: Since the threadpoolexecutor doesn't have a "started" callback, I wouldn't worry about this now.

adlfs/transfer.py
jbcrail: The `ADLTransferClient.submit()` puts each file in a pending state. The files switch to the transferring state when calling `ADLTransferClient.run()`.

martindurant: Exactly - but they are not yet transferring: even though the chunks are submitted to the thread pool, none of them will necessarily get executed until later

jbcrail: This was a tradeoff. I change the states of a file and its chunks when submitting to the thread pool, because it was simpler to handle everything within `ADLTransferClient`. To more precisely record the state of a file or chunk, the transfer/merge callables would need some way to change the state from within. This could be handled in a future PR.

adlfs/transfer.py
martindurant: Agreed, I think this is the best we can do with threads.

martindurant:
When uploading, I sometimes see this:
```
/Users/mdurant/Downloads/azure-data-lake-store-python/adlfs/transfer.py in progress(self)
    214     def progress(self):
    215         files = []
--> 216         for key in self._files:
    217             src, dst = key
    218             chunks = []

AttributeError: 'ADLTransferClient' object has no attribute '_files'
```

i.e., I think the first check is happening before the file uploads start?

martindurant:
i am finding, when uploading a small file with even smaller chunks, that no file is produced at the far end.

jbcrail:
I made a newbie pickle mistake. `__getstate__()` was calling an instance property (`self.progress`), which caused the attribute error.

adlfs/transfer.py
```diff
@@ -333,6 +333,8 @@ def monitor(self, poll=0.1, timeout=0):
 
     def __getstate__(self):
         dic2 = self.__dict__.copy()
+        dic2.pop('_transfer', None)
+        dic2.pop('_merge', None)
         dic2.pop('_pool', None)
         dic2.pop('_shutdown_event', None)
         dic2['_files'] = self._files.copy()
```
martindurant: This is where the attribute error seems to happen

martindurant:
I am getting things like this
```
In [21]: up.client.progress
Out[21]: [File(src='README', dst=AzureDLPath('uploads/READ'), state='merging', nbytes=578, start=1473884538.284629, stop=None, chunks=[Chunk(name='/tmp/eef329887a/part_500', state='finished', offset=500, retries=5), Chunk(name='/tmp/eef2e8427a/part_0', state='finished', offset=0, retries=5)])]
```
but the part_500 doesn't appear on the server - presumably why the subsequent concat then fails. So some things went wrong: one of the files didn't send (don't know why), but the error was not picked up. The subsequent error in merging was noticed, but the upload finished without warning anyway.

jbcrail:
Thanks for the output. I'm now able to reproduce similar behavior, but I don't have a fix yet.

adlfs/transfer.py
begoldsm: I am with Martin on this, it will be good to have more descriptive comments here to help with maintenance in the future.

adlfs/multithread.py
```diff
@@ -14,24 +14,20 @@
...
+from .transfer import ADLTransferClient
 from .utils import commonprefix, datadir, read_block, tokenize
 
 MAXRETRIES = 5
 
 logger = logging.getLogger(__name__)
 
 
-class ADLDownloader:
+class ADLDownloader(object):
```
begoldsm: Is this "object" that is passed in documented below so the caller knows what this object should be?

adlfs/multithread.py
```diff
@@ -93,146 +104,79 @@ def _setup(self):
...
                 if tries >= retries:
-                    logger.debug('Aborting %s, byte offset %s', lfile, offset)
+                    logger.debug('Aborting %s, byte offset %s', dst, offset)
                     raise
-    logger.debug('Downloaded to %s, byte offset %s', lfile, offset)
+    logger.debug('Downloaded to %s, byte offset %s', dst, offset)
 
 
-class ADLUploader:
+class ADLUploader(object):
```
begoldsm: same question. Is this the ADLTransferClient or something else?

adlfs/multithread.py
martindurant: No, this denotes the parent class. It makes no difference in py3, but has subtle implications in the setup and teardown mechanism for instances for older pythons. It's a "can't hurt, might help" situation and doesn't chance the calling signature.

adlfs/transfer.py
```diff
@@ -0,0 +1,368 @@
...
+# license information.
+# --------------------------------------------------------------------------
+
+"""
+High performance multi-threaded module to up/download
+
+Calls method in `core` with thread pool executor to ensure the network
+is used to its maximum throughput.
+
+Only implements upload and download of (massive) files and directory trees.
```
begoldsm: I am not sure I understand what this means. Does this mean it can't upload a small file or directory?

adlfs/transfer.py
```diff
@@ -0,0 +1,368 @@
...
+import multiprocessing
+import os
+import pickle
+import threading
+import time
+import uuid
+
+from adlfs.utils import clamp
+
+MAX_THREADS = 64
```
begoldsm: Why is this the max? Can we not handle more than 64 threads? Does this guarantee maximum throughput on any machine (say, a machine with a 10gbps nic, 64 cores and 128gb of memory)?

adlfs/transfer.py
```diff
@@ -0,0 +1,368 @@
...
+    ----------
+    adlfs: ADL filesystem instance
+    name: str
+        Unique ID used for persistence.
+    transfer: callable
+        Function or callable object invoked when transferring chunks. See
+        ``Function Signatures``.
+    merge: callable [None]
+        Function or callable object invoked when merging chunks. If None,
+        then merging is skipped. See ``Function Signatures``.
```
begoldsm: So this would happen in the case where we are uploading a lot of small files or when the file we are uploading is small enough to not have chunks?

adlfs/transfer.py
```diff
@@ -0,0 +1,368 @@
...
+    nthreads: int [None]
+        Number of threads to use (range is [1, MAX_THREADS]). If None, uses the
+        number of cores.
```
begoldsm: what if a machine has more than 64 cores?

adlfs/transfer.py
```diff
@@ -0,0 +1,368 @@
...
+    chunksize: int [2**26]
+        Number of bytes in each chunk for splitting big files. Files smaller
+        than this number will always be transferred in a single thread.
```
begoldsm: segment size per file  for upload should be 256mb, and uploads within each segment should be in 4mb increments. This is critical to prevent file fragmentation on the server.

adlfs/transfer.py
```diff
@@ -0,0 +1,368 @@
...
+    tmp_path: str ['/tmp']
+        Path used for temporarily storing transferred chunks until chunks
+        are gathered into a single file. If None, then each chunk will be
+        written into the same file.
```
begoldsm: This path should be unique for each upload to make it easier for the caller to clean up failed uploads of specific files. Otherwise they have to guess which files in this tmp folder correspond to which target files, which is painful and almost impossible. Each upload should get its own "temporary segment folder" which you clean up at the end by calling concat with deletesourcedirectory=true.

adlfs/transfer.py
```diff
@@ -0,0 +1,368 @@
...
+    tmp_prefix: str ['part_']
+        If given and not None, this is used to provide a prefix to the
+        temporary chunk.
+    tmp_unique: bool [True]
+        If True, then a unique ID will be generated to create a subdirectory
+        containing the temporary chunks. Otherwise, all temporary chunks
+        will be placed in `tmp_path`.
```
begoldsm: Ah, I see. Can you give an example of what this looks like for a given file? Will the caller be able to identify what file this unique tmp corresponds to? It should probably also include the target file name somewhere in the path for identification by the caller in the event of failure.

adlfs/transfer.py
```diff
@@ -0,0 +1,368 @@
...
+    persist_path: str [None]
+        Path used for persisting a client's state. If None, then `save()`
+        and `load()` will be empty operations.
```
begoldsm: Is there a way to have this default to some temporary location? That is probably difficult in a cross platform environment, but reducing the number of parameters the caller needs to worry about is ideal. This is not critical, though.

adlfs/transfer.py
```diff
@@ -0,0 +1,368 @@
...
+    to listen for this.
+
+    See Also
+    --------
+    adlfs.multithread.ADLDownloader
+    adlfs.multithread.ADLUploader
+    """
+
+    def __init__(self, adlfs, name, transfer, merge=None, nthreads=None,
+                 chunksize=2**26, tmp_path='/tmp', tmp_prefix='part_',
```
begoldsm: >2**26 [](start = 27, length = 5)

should be 256mb (2**28)

adlfs/transfer.py
begoldsm: Agree. Not sure if we need to track total time in here (since that is easy for someone to measure externally) but the important things to track are:
Total File Size (in bytes)
Bytes Uploaded
Total chunks
Chunks remaining.

---
In reply to: [78791135](https://github.com/Azure/azure-data-lake-store-python/pull/64#discussion_r78791135) [](ancestors = 78791135)

adlfs/transfer.py
begoldsm: Curious: are we tracking these times for debugging/instrumentation purposes or are they used somewhere else? For debugging and instrumentation I am all for it, just want to make sure I understand the reasoning.

---
In reply to: [78791538](https://github.com/Azure/azure-data-lake-store-python/pull/64#discussion_r78791538) [](ancestors = 78791538)

tests/test_transfer.py
```diff
@@ -0,0 +1,30 @@
...
+# --------------------------------------------------------------------------
+
+import pytest
+import time
+
+from tests.testing import azure
+from adlfs.transfer import ADLTransferClient
+
+
+@pytest.mark.skipif(True, reason="skip until resolve timing issue")
```
begoldsm: Do you plan on fixing this test prior to check-in, or in a future PR?

begoldsm:
:clock1:

adlfs/transfer.py
jbcrail: This docstring was copied from the `multithread` module and doesn't make sense here. I've updated it to be more descriptive.

adlfs/transfer.py
jbcrail: The initial rationale to limit the threads was based on performance tests. If a more capable machine is available, the max optimal value could be higher, so I've removed the upper bound.

adlfs/transfer.py
jbcrail: In the case where we are transferring only one small file or one chunk, then we disable the merge step and do not write to a temporary file.

adlfs/transfer.py
jbcrail: I removed the upper bound.

adlfs/transfer.py
jbcrail: I updated the default.

tests/test_transfer.py
jbcrail: This can be address in a future PR. This particular test is a similar version once contained in the `multithread` module.

adlfs/transfer.py
jbcrail: I added a `temporary_path` property to give access to the value. I also removed the `tmp_path` parameter in lieu of using the target basename instead. I documented the full temporary path used by the client.

adlfs/transfer.py
martindurant: This was my suggestion. My thinking was, a user can always change a module-level constant like this at runtime (without changing the code), so it's a _default_ _default_.

adlfs/transfer.py
martindurant: Would it be better to say: "for each file containing only one chunk, no merge function will be called, even if provided" - but you can have many-chunk files and single-chunk files as part of the same transfer?

adlfs/transfer.py
martindurant: The best value for downloads (not uploads) was found to be smaller than 256MB (which may impose memory problems), that's where the number came from. Whether the default of the combined transfer class should be that of the upload or download shouldn't matter, since each subclass has its own default too. Maybe there should be no default here at all.

(NB: uploading of files is always in 4MB chunks or smaller, but this logic is in the underlying ADLFile class)

adlfs/transfer.py
jbcrail: I liked your description and added it to the docs.

adlfs/transfer.py
begoldsm: As long as it is obvious to a user/caller how to get a higher value (and easy in a CLI implementation) then I am ok with having an internal default, I just want to make sure it is well documented so users understand what the intention is. As it was a user would think that adding more threads wouldn't give value or would not work, potentially.

---
In reply to: [78975720](https://github.com/Azure/azure-data-lake-store-python/pull/64#discussion_r78975720) [](ancestors = 78975720)

PR #68, opened 2016-09-16
Improve download performance

jbcrail:
Instead of a thread writing a chunk in one call, we now break the chunk into blocks and write each block until finished. The default chunk size (256MB) and block size (32MB) is consistent across uploads and downloads.

I also added a `blocksize` parameter so that the user could tune as necessary.


begoldsm:
:shipit:

begoldsm:
@jbcrail can you sync real quick and update?

PR #70, opened 2016-09-19
Change default size of download write block

jbcrail:
This fixes a regression in the downloader. A 4MB block was determined to be the optimal size.


jbcrail:
```
Download 1 100GB: 1305.0513s
Download 100 1GB: 1345.8531s
```

begoldsm:
@jbcrail thank you!

Issue #73, opened 2016-09-20
Place temporary files under destination path


jbcrail:
As mentioned by Ben, we should place uploaded chunks in a temporary directory under the destination path. The proposed temporary directory/files are in keeping with other SDK implementations.

If the destination file is `/user/deeperFolder/bar.txt`, then the temporary directory will be:

```
/user/deeperFolder/bar.txt.segments.{random string}/
```

Within the temporary directory, the temporary chunk filenames will be:

```
bar.txt.{random string}.segment{int}
```

where `int` is the chunk number.

PR #74, opened 2016-09-20
PR: Upload/download 100GB fails without errors

### LINKED ISSUE #71: Pri0: Upload and download of 100GB file fails with no error, debug logs do not indicate what the error is either.
Attached is the debug output (below). This blocks my ability to run upload and download tests.
Additionally, it is not, by default, creating the tmp files closer to the target file. It is instead still creating them inside of /tmp, which makes it difficult for me to know which segments are associated with which file upload attempt.

[outlogUpload.txt](https://github.com/Azure/azure-data-lake-store-python/files/481450/outlogUpload.txt)
[outlogDownload.txt](https://github.com/Azure/azure-data-lake-store-python/files/481457/outlogDownload.txt)

jbcrail:
The temporary file path still uses `/tmp` until #73 is resolved.

Fixes #71

### DIFF
--- a/adlfs/core.py
+++ b/adlfs/core.py
@@ -719,7 +719,7 @@ def _fetch_range(rest, path, start, end, max_attempts=10):
                              read='true')
             return resp
         except Exception as e:
-            logger.debug('Exception %e on ADL download, retrying', e,
+            logger.debug('Exception %s on ADL download, retrying', e,
                          exc_info=True)
     raise RuntimeError("Max number of ADL retries exceeded")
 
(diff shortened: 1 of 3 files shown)


jbcrail:
All runs completed without error:

```
Upload 1 100GB: 1021.7922s
Upload 100 1GB: 1004.3417s

Download 1 100GB: 1212.6360s
Download 100 1GB: 2191.3073s
```

begoldsm:
@jbcrail what thread counts are you using? I am glad that they are all passing, just curious why we are seeing such long runtimes.

PR #76, opened 2016-09-21
PR: Metadata improvements

### LINKED ISSUE #66: PRI 0: Use the Offset parameter in append to ensure we are always appending at the offset we expect
The append API allows for a query parameter called offset, which allows you to specify where you think the end of the file is to append data to. This is very useful for us, since it allows us to submit an append call where we "think" the file should currently be. If data has already been uploaded at that location, you will receive a 400 error with a BadOffsetException, which you can catch. This will indicate that data is already at that location and we can move on to the next 4mb chunk to upload. This will help preserve the uploaded file and ensure that its final length is the same as the source file.

Here is a sample request/response where the offset is not the end of the file (meaning there is already data at or ahead of the offset specified):

**Request:** 
POST https://adlspysample01.azuredatalakestore.net/webhdfs/v1/sample.txt?op=APPEND&append=true&**offset=4**&api-version=2015-10-01-preview HTTP/1.1
x-ms-client-request-id: 33401497-4ac0-451e-ab13-eccd305e3706
accept-language: en-US
Authorization:
User-Agent: Microsoft.Azure.Management.DataLake.Store.DataLakeStoreFileSystemManagementClient/0.12.6-preview AzurePowershell/v2.1.0.0
CommandName: Add-AzureRmDataLakeStoreItemContent
ParameterSetName: __AllParameterSets
Content-Type: application/octet-stream
Host: adlspysample01.azuredatalakestore.net
Content-Length: 4
Expect: 100-continue

test

**Response**
HTTP/1.1 400 Bad Request
Cache-Control: no-cache
Pragma: no-cache
Content-Length: 260
Expires: -1
x-ms-request-id: a8b65992-21f3-4c5d-8fe0-f0b49d0320f1
Server-Perf: [a8b65992-21f3-4c5d-8fe0-f0b49d0320f1][ AuthTime::0::PostAuthTime::0 ][S-FsOpenStream :: 00:00:010 ms]%0a[BufferingTime :: 00:00:000 ms]%0a[WriteTime :: 00:00:000 ms]%0a[S-FsAppendStream :: 00:00:036 ms]%0a[S-FsCloseHandle :: 00:00:001 ms]%0a[APPEND :: 00:00:269 ms]%0a
Status: 0x83090015
Date: Thu, 15 Sep 2016 01:21:15 GMT

**{"RemoteException":{"exception":"BadOffsetException","message":"FsAppendStream failed with error 0x83090015 (Bad offset). [a8b65992-21f3-4c5d-8fe0-f0b49d0320f1][2016-09-14T18:21:16.0457431-07:00]","javaClassName":"org.apache.hadoop.fs.adl.BadOffsetException"}}**

jbcrail:
The main purpose of this PR was to add the necessary file/chunk metadata to make it easier for testing, monitoring progress, and verifying the completed files/chunks. I also refactored how we update the file/chunk statuses while in progress (we use callbacks now instead of polling).

This will allow use to add pre- and post-checks after every chunk transfer.

Fixes #66.

### DIFF
--- a/adlfs/core.py
+++ b/adlfs/core.py
@@ -735,7 +735,7 @@ def _fetch_range(rest, path, start, end, max_attempts=10):
 
 
 def _put_data(rest, op, path, data, max_attempts=10, **kwargs):
-    logger.debug("Put: %s, %s", path, kwargs)
+    logger.debug("Put: %s %s, %s", op, path, kwargs)
     for i in range(max_attempts):
         try:
             resp = rest.call(op, path=path, data=data, **kwargs)
--- a/adlfs/core.py
+++ b/adlfs/core.py
@@ -619,10 +619,12 @@ class AzureDLFile(object):
         if self.mode in {'wb', 'ab'} and not self.closed:
             if self.buffer.tell() == 0:
                 if force and self.first_write:
-                    self.azure.azure.call('CREATE',
-                                          path=self.path.as_posix(),
-                                          overwrite='true',
-                                          write='true')
+                    _put_data(self.azure.azure,
+                              'CREATE',
+                              path=self.path.as_posix(),
+                              data=None,
+                              overwrite='true',
+                              write='true')
                     self.first_write = False
                 return
             self.buffer.seek(0)
@@ -636,17 +638,19 @@ class AzureDLFile(object):
                     else:
                         limit = place + len(self.delimiter)
                     if self.first_write:
-                        out = self.azure.azure.call('CREATE',
-                                                    path=self.path.as_posix(),
-                                                    data=data[:limit],
-                                                    overwrite='true',
-                                                    write='true')
+                        out = _put_data(self.azure.azure,
+                                        'CREATE',
+                                        path=self.path.as_posix(),
+                                        data=data[:limit],
+                                        overwrite='true',
+                                        write='true')
                         self.first_write = False
                     else:
-                        out = self.azure.azure.call('APPEND',
-                                                    path=self.path.as_posix(),
-                                                    data=data[:limit],
-                                                    append='true')
+                        out = _put_data(self.azure.azure,
+                                        'APPEND',
+                                        path=self.path.as_posix(),
+                                        data=data[:limit],
+                                        append='true')
                     logger.debug('Wrote %d bytes to %s' % (limit, self))
                     data = data[limit:]
                 self.buffer = io.BytesIO(data)
@@ -658,18 +662,20 @@ class AzureDLFile(object):
                     offset = zero_offset + o
                     d2 = data[o:o+self.blocksize]
                     if self.first_write:
-                        out = self.azure.azure.call('CREATE',
-                                                    path=self.path.as_posix(),
-                                                    data=d2,
-                                                    overwrite='true',
-                                                    write='true')
+                        out = _put_data(self.azure.azure,
+                                        'CREATE',
+                                        path=self.path.as_posix(),
+                                        data=d2,
+                                        overwrite='true',
+                                        write='true')
                         self.first_write = False
                     else:
-                        out = self.azure.azure.call('APPEND',
-                                                    path=self.path.as_posix(),
-                                                    data=d2,
-                                                    offset=offset,
-                                                    append='true')
+                        out = _put_data(self.azure.azure,
+                                        'APPEND',
+                                        path=self.path.as_posix(),
+                                        data=d2,
+                                        offset=offset,
+                                        append='true')
                     logger.debug('Wrote %d bytes to %s' % (len(d2), self))
                 self.buffer = io.BytesIO()
             return out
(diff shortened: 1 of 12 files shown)


jbcrail:
All benchmarks completed without errors:
```
[bench_upload_1_100gb] starting...
[bench_upload_1_100gb] finished in 612.4005s
[bench_download_1_100gb] starting...
[bench_download_1_100gb] finished in 1234.5771s
[bench_upload_100_1gb] starting...
[bench_upload_100_1gb] finished in 762.1828s
[bench_download_100_1gb] starting...
[bench_download_100_1gb] finished in 1207.7382s
```

adlfs/multithread.py
```diff
@@ -135,7 +135,7 @@ def touch(self, src, dst):
             with open(dst, 'wb'):
                 pass
 
-        self.client.run(nthreads, monitor, before_scatter=touch)
+        self.client.run(nthreads, monitor, before_start=touch)
```
martindurant: Concatenates file to zero - results in one extra REST call per remote final file, correct? Checking for the existence of the file upfront would be one REST call or zero, if the file list was already cached.

btw: please check, is `adl.invalidate_cache(rfile)` being called when a file upload is finalised?

adlfs/multithread.py
```diff
@@ -183,6 +185,7 @@ def get_chunk(adlfs, src, dst, offset, size, blocksize, retries=MAXRETRIES,
                             logger.debug('Aborting %s, byte offset %s', dst, o)
                             raise
```
martindurant: Return `False`? I thought we wanted to avoid passing back the exception, which might be keeping the stack live following an exception. That would need a check for `False` at the far end, rather than checking the error status.

adlfs/multithread.py
```diff
@@ -327,7 +332,7 @@ def put_chunk(adlfs, src, dst, offset, size, blocksize, retries=MAXRETRIES,
                             logger.debug('Aborting %s, byte offset %s', src, o)
                             raise
```
martindurant: as above

adlfs/transfer.py
```diff
@@ -267,110 +310,101 @@ def progress(self):
...
+                self._fstates[parent] = 'errored'
+        elif future in self._ffutures:
+            src, dst = self._ffutures[future]
+
+            if future.cancelled():
+                self._fstates[(src, dst)] = 'cancelled'
+            elif future.exception():
+                self._fstates[(src, dst)] = 'errored'
+            else:
+                result = future.result()
```
martindurant: Should we not check the number of bytes here?
Also, here is where we would check for False (or the string repr of the exception, perhaps) in case of exception caught in the thread and not passed back.

adlfs/transfer.py
```diff
@@ -267,110 +310,101 @@ def progress(self):
...
+    def run(self, nthreads=None, monitor=True, before_start=None):
         self._nthreads = nthreads or self._nthreads
         for src, dst in self._files:
-            self._files[(src, dst)]['start'] = time.time()
-            self._fstates[(src, dst)] = 'transferring'
-            if before_scatter:
-                before_scatter(self._adlfs, src, dst)
-            self._scatter(src, dst, self._transfer)
+            if before_start:
+                before_start(self._adlfs, src, dst)
```
martindurant: If this is `touch()`, how does the `src` parameter end up getting dropped here?

jbcrail: This is a local function, not `adlfs.touch`. I pass both parameters to be generic. These two lines were included when creating the generic client, but they may be removed at some point.

adlfs/multithread.py
jbcrail: I now return the number of bytes written and the exception as a string (None if not present).

adlfs/multithread.py
jbcrail: Same as above

adlfs/transfer.py
jbcrail: You're right. I just added this check.

adlfs/multithread.py
jbcrail: `invalidate_cache` is not called upon finalization.

We may have to hold off on the file existence check. I talked with Ben about the conditions for forcing an overwrite. I should create a separate issue to track this.

martindurant: It should be called, else any `info` to check the file size might get not-found or else previous size.

jbcrail:
Updated benchmark results:
```
[bench_upload_1_50gb] starting...
local file      : D:\ingress\50gbfile.txt
local file size : 52425000000
remote file     : joseph/50gbfile.txt
remote file size: 52425000000
[bench_upload_1_50gb] finished in 263.0346s

[bench_download_1_50gb] starting...
local file      : D:\ingress\50gbfile.txt.out
local file size : 52425000000
remote file     : joseph/50gbfile.txt
remote file size: 52425000000
[bench_download_1_50gb] finished in 397.7892s

>>> md5sum('D:\\ingress\\50gbfile.txt')
'60f5fba5f5eb30ab75f84214e184acae'
>>> md5sum('D:\\ingress\\50gbfile.txt.out')
'6c37bf52d23560df955d36b6d32b822a'
```

adlfs/core.py
```diff
@@ -728,6 +734,18 @@ def _fetch_range(rest, path, start, end, max_attempts=10):
     raise RuntimeError("Max number of ADL retries exceeded")
 
 
+def _put_data(rest, op, path, data, max_attempts=10, **kwargs):
+    logger.debug("Put: %s, %s", path, kwargs)
```
martindurant: Should include `op`

jbcrail:
Updated benchmark results:
```
md5sum('D:\ingress\50gbfile.txt') = 60f5fba5f5eb30ab75f84214e184acae
[bench_upload_1_50gb] starting...
local file      : D:\ingress\50gbfile.txt
local file size : 52425000000
remote file     : joseph/50gbfile.txt
remote file size: 52425000000
[bench_upload_1_50gb] finished in 281.7676s
[bench_download_1_50gb] starting...
local file      : D:\ingress\50gbfile.txt.out
local file size : 52425000000
remote file     : joseph/50gbfile.txt
remote file size: 52425000000
[bench_download_1_50gb] finished in 466.3828s
md5sum('D:\ingress\50gbfile.txt.out') = 60f5fba5f5eb30ab75f84214e184acae
```

begoldsm:
Thank you for the updated benchmark test @jbcrail!

PR #81, opened 2016-09-28
PR: Move temporary files under destination path

### LINKED ISSUE #73: Place temporary files under destination path

jbcrail:
Fixes #73.

### DIFF
--- a/adlfs/transfer.py
+++ b/adlfs/transfer.py
@@ -136,14 +136,13 @@ class ADLTransferClient(object):
     blocksize: int [2**25]
         Number of bytes for a block. Within each chunk, we write a smaller
         block for each API call. This block cannot be bigger than a chunk.
-    tmp_path: str ['/tmp']
-        Path used for temporarily storing transferred chunks until chunks
-        are gathered into a single file. If None, then each chunk will be
-        written into the same file.
-    tmp_unique: bool [True]
-        If True, then a unique ID will be generated to create a subdirectory
-        containing the temporary chunks. Otherwise, all temporary chunks
-        will be placed in `tmp_path`.
+    chunked: bool [True]
+        If set, each transferred chunk is stored in a separate file until
+        chunks are gathered into a single file. Otherwise, each chunk will be
+        written into the same destination file.
+    unique_temporary: bool [True]
+        If set, transferred chunks are written into a unique temporary
+        directory.
     persist_path: str [None]
         Path used for persisting a client's state. If None, then `save()`
         and `load()` will be empty operations.
@@ -155,11 +154,10 @@ class ADLTransferClient(object):
     ---------------
 
     When a merge step is available, the client will write chunks to temporary
-    files before merging. The exact temporary file is dependent upon on two
-    parameters (`tmp_path`, `tmp_unique`). Given those values, the full
-    temporary file looks like this in pseudo-BNF:
+    files before merging. The exact temporary file looks like this in
+    pseudo-BNF:
 
-    >>> # /{tmp_path}[/{unique_str}]/{basename}_{offset}
+    >>> # {dirname}/{basename}.segments[.{unique_str}]/{basename}_{offset}
 
     Function Signatures
     -------------------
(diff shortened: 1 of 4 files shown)


jbcrail:
All benchmarks completed without errors (64 threads):
```
[bench_upload_1_50gb] starting...
local file      : D:\ingress\50gbfile.txt
local file size : 52425000000
remote file     : /joseph/50gbfile.txt
remote file size: 52425000000
[bench_upload_1_50gb] finished in 277.1660s

[bench_download_1_50gb] starting...
local file      : D:\ingress\50gbfile.txt.out
local file size : 52425000000
remote file     : /joseph/50gbfile.txt
remote file size: 52425000000
[bench_download_1_50gb] finished in 455.2181s

60f5fba5f5eb30ab75f84214e184acae D:\ingress\50gbfile.txt
60f5fba5f5eb30ab75f84214e184acae D:\ingress\50gbfile.txt.out

[bench_upload_50_1gb] starting...
local file      : D:\ingress\50_1GB_Files
local file size : 52425000000
remote file     : /joseph/50_1GB_Files
remote file size: 52425000000
[bench_upload_50_1gb] finished in 322.3616s

[bench_download_50_1gb] starting...
local file      : D:\ingress\50_1GB_Files.out
local file size : 52425000000
remote file     : /joseph/50_1GB_Files
remote file size: 52425000000
[bench_download_50_1gb] finished in 419.6399s

16ca4ecebb7987748097be6a87a98c20 D:\ingress\50_1GB_Files
16ca4ecebb7987748097be6a87a98c20 D:\ingress\50_1GB_Files.out
```

adlfs/transfer.py
```diff
@@ -256,18 +253,17 @@ def submit(self, src, dst, length):
             'pending', 'running', 'finished', 'cancelled', 'errored')
 
         # Create unique temporary directory for each file
-        if self._tmp_unique and self._tmp_path:
-            tmpdir = os.path.join(self._tmp_path, uuid.uuid4().hex)
+        if self._chunked and self._unique_temporary:
+            tmpdir = dst.parent / "{}.segments.{}".format(dst.name, self._unique_str)
```
martindurant: Are these files in the same directory as the target file or a directory under it?
If the latter, do we need to create that directory, when downloading?

adlfs/transfer.py
```diff
@@ -217,11 +215,9 @@ class ADLTransferClient(object):
     adlfs.multithread.ADLUploader
     """
 
-    DEFAULT_TMP_PATH = 'tmp'
-
     def __init__(self, adlfs, name, transfer, merge=None, nthreads=None,
-                 chunksize=2**28, blocksize=2**25, tmp_path=DEFAULT_TMP_PATH,
-                 tmp_unique=True, persist_path=None, delimiter=None):
+                 chunksize=2**28, blocksize=2**25, chunked=True,
```
martindurant: Do we expect the defaults ever not to be used? Or is this just included here for testing?

begoldsm: I think it is ok to include the ability to tweak these things, but yes, we should default everything to "optimal" performance and usability from our perspective. That way, if a user wants to change it, its assumed they know what they are doing and have a good reason.

adlfs/transfer.py
begoldsm: Agree with @martindurant here. For download the directory will need to be created if it doesn't exist first, right? For upload this is not required since our service will create the entire path to the file automatically.

adlfs/transfer.py
jbcrail: I believe all parameters except `unique_temporary`, which is used for testing, are overridden by `ADLDownloader` and `ADLUploader`, so the defaults could be removed.

But, like @begoldsm mentioned, I thought optimal defaults were needed. We allow an advanced user to override these values by creating an `ADLTransferClient` object and passing it into `ADLDownloader` or `ADLUploader`. Removing the defaults would force the user to set each parameter.

adlfs/transfer.py
begoldsm: ok great, thanks for the confirmation @jbcrail

PR #82, opened 2016-09-28
PR: Force overwrite

### LINKED ISSUE #53: ADLUploader defaults to overwrite if file exists
The operation should fail and the user should be told that they must explicitly use a force flag to overwrite an existing file when using ADLUploader. Today I can re-run the same upload command to the same target location and it won't complain.

jbcrail:
Fixes #53.


adlfs/cli.py
```diff
@@ -305,10 +310,12 @@ def do_put(self, line):
         parser.add_argument('remote_path', type=str, nargs='?', default='.')
         parser.add_argument('-b', '--chunksize', type=int, default=2**28)
         parser.add_argument('-c', '--threads', type=int, default=None)
+        parser.add_argument('-f', '--force', action='store_true')
```
begoldsm: Does there need to be a default here? or will it automatically default to false?

jbcrail: No default needed. It automatically sets to false by default. `store_true` is the action taken when the `force` flag is present.

adlfs/multithread.py
```diff
@@ -57,13 +58,17 @@ class ADLDownloader(object):
         by constructor.
     run: bool [True]
         Whether to begin executing immediately.
+    overwrite: bool [False]
+        Whether to forcibly overwrite existing files/directories.
```
martindurant: Please extend: if it's a directory, will quit if this is False, whether any files would be overwritten or not - only matching filenames are actually overwritten if True.

martindurant:
+1

PR #85, opened 2016-09-29
Skip existing block

### LINKED ISSUE #66: PRI 0: Use the Offset parameter in append to ensure we are always appending at the offset we expect

martindurant:
Fixes #66


adlfs/core.py
```diff
@@ -743,6 +743,9 @@ def _put_data(rest, op, path, data, max_attempts=10, **kwargs):
         except (PermissionError, FileNotFoundError):
             raise
         except Exception as e:
+            if '"BadOffsetException"' in repr(e):
```
begoldsm: @martindurant is it possible for this check to be more targeted to just the uploader logic? Callers should be able to decide how they want to handle badOffsetExceptions when they build their own apps that use the SDK, we don't want to swallow this exception for all cases, just in our implementation of the uploader where we know with near certainty what is going on.

martindurant: The two modes for opening a file via the ADLFileSystem will never allow this situation to occur (you either clobber the file to zero, or append at the end; there is no seek in write mode).
Putting the except in the uploader is more tricky, because first this function would need to retry the requisite number of times before control is passed back.

martindurant: (but we could rename the `loc` attribute to `_loc` to indicate that users shouldn't mess with it)

begoldsm: @martindurant I think we are missing the purpose of including the offset parameter in the first place. It is strictly forbidden to append at any offset other than the very end of the file. You will get an exception if you try to append:
a) an index where there is already data
b) at an index past the end of the file.

The purpose of utilizing the offset parameter is to protect the multi-threaded upload portion from inadvertently uploading too much data without just failing out. Here is the specific scenario we are trying to protect against:
1. multi-part uploader begins upload 1GB file
2. while appending a 4mb block the front end doesn't respond back, so we cancel the connection and start the retry logic
3. What we don't know is that the append succeeded in the back end. The data is there and in the file.
4. If we append now (as a retry) without including the offset we expect the file to be at, then the append will succeed (again) and we will get an extra 4mb of data in the file.

The solution for this particular problem, is that applications that use the client can protect themselves from this scenario by explicitly specifying the offset they want, then making a decision about what to do if they get a badoffset exception.

The client itself should only give the caller options, it should not make any decisions for the caller directly.

azure/datalake/store/core.py
```diff
@@ -748,6 +748,13 @@ def _put_data(rest, op, path, data, max_attempts=10, **kwargs):
         except (PermissionError, FileNotFoundError) as e:
             rest.log_response_and_raise(resp, e)
         except Exception as e:
+            if '"BadOffsetException"' in repr(e):
```
begoldsm: Again, this isn't quite right. In our Multithreaded upload implementation, we have reasonable confidence that we will always be either at the right offset (end of file) or behind the end of file (append already succeeded). But in the core logic, someone else could use our "append" implementation and incorrectly try to append past the end of the file. That will also result in a BadOffsetException, but if they retry we will silently continue, even though their data is not getting committed (since it is beyond the end of the file).

The right place to do this logic is inside of our Multithread.py implementation, where we have confidence in our offset management and can be explicit about how to handle badoffsetexceptions. We should not make a general statement of how badoffsetexceptions should be handled for all custom clients that use our core.py, since that can be surprising behavior.

begoldsm:
@jbcrail were you able to verify that this resolved this issue on the repro machine?

jbcrail:
The results are uneven. On one VM, the results were consistently good. But on your VM, the results were inconsistent.

PR #87, opened 2016-09-30
Produce textual feedback on transfer process

martindurant:
(controlled using verbose= flag in constructor)

### DIFF
--- a/adlfs/multithread.py
+++ b/adlfs/multithread.py
@@ -69,9 +69,10 @@ class ADLDownloader(object):
     adlfs.transfer.ADLTransferClient
     """
     def __init__(self, adlfs, rpath, lpath, nthreads=None, chunksize=2**28,
-                 blocksize=2**22, client=None, run=True, overwrite=False):
+                 blocksize=2**22, client=None, run=True, overwrite=False,
+                 verbose=True):
         if not overwrite and adlfs.exists(rpath):
-            raise FileExistsError(rpath)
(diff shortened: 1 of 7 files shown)


adlfs/multithread.py
```diff
@@ -69,9 +69,10 @@ class ADLDownloader(object):
     adlfs.transfer.ADLTransferClient
     """
     def __init__(self, adlfs, rpath, lpath, nthreads=None, chunksize=2**28,
-                 blocksize=2**22, client=None, run=True, overwrite=False):
+                 blocksize=2**22, client=None, run=True, overwrite=False,
+                 verbose=True):
         if not overwrite and adlfs.exists(rpath):
```
begoldsm: I am confused by this logic. Shouldn't it be checking for the existence of the local file? Why are the checks inverted for upload and download based on the errors being raised.

martindurant: We had it the wrong way around - switched in this PR

begoldsm:
@martindurant is there a test for overwrite= false that validates the right path is being returned in the `FileExistsError`? That should be added here as a regression test since that was the bug previously. Also that it properly fails when the right file already exists, since before it complained when the source file existed.

martindurant:
See tests/test_multithread.py , test_upload_overwrite, test_download_overwrite

begoldsm:
Great, thanks @martindurant! I will hold off on the merge until that test is added to this PR.

PR #90, opened 2016-10-03
Persist run

martindurant:
Hooks for persisting and resuming up/downloads.

Also allows the same from the CLI.


begoldsm:
@martindurant can you include an example of how this resume functionality would be used? For example, does the caller need to specify anything in particular on the re-execution? How does it know it needs to do a resume instead of a fresh run, etc.

martindurant:
I will give a precis here, but this should be described in full in the docs.

Every up/download saves itself on each change in state (i.e., any chunk completed/errored). They will remain amongst the list of persisted transfers until they are completed, unless the user explicitly calls clear. Any future session (or later in the same session) can resume a transfer, and all chunks that were not already successfully moved will get restarted, until done.

In the CLI, this manifests as list_uploads/list_downloads, which will give a text list of hash keys for saved transfers, and brief status that was last recorded. We trust that the state of the source and destination hasn't changed (i.e., this is not meant to be resumed weeks later). 

In the library, the user could in theory recreate a transfer from start by calling `setup()` again, but I don't see the use case, as it's as easy to call the constructor.

begoldsm:
Ok great. thanks for the crisp explanation @martindurant. You mentioned it being described in full in the docs. Is that something you will add to this change or that we need to track separately?

martindurant:
The docs for the CLI is extremely limited; would it be best to pass on my comments above to your docs team when they get around to fleshing it out? 

Perhaps the text as I have it should be put into docstrings and the auto-help for the CLI. @jbcrail , do you plan to spend any time on docs still?

begoldsm:
@martindurant oh I didn't mean for the CLI. I meant for the library itself. Do we need to have similar documentation to what you described for users of the client itself (not the CLI) to understand how to properly resume on a future run?

martindurant:
Oh OK, I follow you. In that case I will insert this text into doc strings, and they'll show up automatically in the API pages. 
The CLI, which is presumably aimed at more novice users, I feel could use a much higher-level description, and that is what is missing.
I see this branch no longer merges clearly, so I'll do nice doc strings when I sort that out.

begoldsm:
@martindurant great thank you!

PR #93, opened 2016-10-04
Optimize the merge call to remove the temp dir

begoldsm:
This results in the concat operation taking substantially less time.


begoldsm:
@jbcrail can you take a look and make sure there aren't other considerations that need to be made if the segment directory is getting deleted as part of merge (for example, removal of explicit deletions of the segment directory)?

jbcrail:
@begoldsm This looks good. We don't delete the temporary directory elsewhere, so this shouldn't have any side effects. My only comment would be to use the named parameter, `delete_source=True`, for clarity.

PR #94, opened 2016-10-05
PR: Actionable errors

### LINKED ISSUE #91: Client Usability: Error messages must be actionable and include trace IDs from the server
When the user receives an error from the client it must be actionable. This means it either
1. Tells the user what they did wrong and how they can correct (FileNotFound, AccessDenied, Conflict, etc).
2. Gives the user enough information to effectively escalate to service engineers (Full error responses from the service, x-ms-client-request-id, traceId, etc.).

Right now, in the parallel upload/download client we are returning a "RuntimeError: Max number of ADL retries exceeded" message. This should be updated to also include what the error was that kept resulting in retries so the user knows what the underlying failure was that caused us to abort the run.

This should also be true for all single calls, since this is a service and it could return a 500 error at some point (due to service unavailability for example) and its important that customers can give the service engineers as much information as possible to quickly and correctly identify and resolve their issue.

jbcrail:
When debugging, the request and response is grouped respectively into a single message for clearer logs. We also now record Azure operation and parameters.

When an exception occurs, we record the response and body along with the exception. This will happen with all HTTP status codes in the 4xx or 5xx range.

I also removed unused imports and file/chunk metadata.

Fixes #91.

### DIFF
--- a/azure/datalake/store/lib.py
+++ b/azure/datalake/store/lib.py
@@ -216,18 +216,34 @@ class DatalakeRESTInterface:
             self.token = refresh_token(self.token)
             self.head = {'Authorization': 'Bearer ' + self.token['access']}
 
-    def _log_request(self, method, url, headers):
-        logger.debug("Request Method: %s", method.upper())
-        logger.debug("Request URL: %s", url)
-        for header in headers:
-            logger.debug("Request Header: %s='%s'", header, headers[header])
+    def _log_request(self, method, url, op, path, params, headers):
+        msg = "HTTP Request\n{} {}\n".format(method.upper(), url)
+        msg += "{} '{}' {}\n\n".format(
+            op, path,
+            " ".join(["{}={}".format(key, params[key]) for key in params]))
+        msg += "\n".join(["{}: {}".format(header, headers[header])
+                          for header in headers])
+        logger.debug(msg)
 
     def _log_response(self, response, payload=False):
-        logger.debug("Response Status Code: %s", response.status_code)
-        for header in response.headers:
-            logger.debug("Response Header: %s='%s'", header, response.headers[header])
+        msg = "HTTP Response\n{}\n{}\n".format(
+            response.status_code,
+            "\n".join(["{}: {}".format(header, response.headers[header])
+                       for header in response.headers]))
         if payload:
-            logger.debug("Response Body: %s", pprint.pformat(response.content))
+            msg += "\n{}".format(pprint.pformat(response.content))
+        logger.debug(msg)
+
+    def log_response_and_raise(self, response, exception):
+        msg = "Exception " + repr(exception)
+        if response:
+            msg += "\n{}\n{}\n\n{}".format(
+                response.status_code,
+                "\n".join(["{}: {}".format(header, response.headers[header])
+                           for header in response.headers]),
+                pprint.pformat(response.content))
+        logger.error(msg)
+        raise exception
 
     def call(self, op, path='', **kwargs):
         """ Execute a REST call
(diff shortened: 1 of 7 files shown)


azure/datalake/store/lib.py
```diff
@@ -216,18 +216,34 @@ def _check_token(self):
...
-        logger.debug("Response Status Code: %s", response.status_code)
-        for header in response.headers:
-            logger.debug("Response Header: %s='%s'", header, response.headers[header])
+        msg = "HTTP Response\n{}\n{}\n".format(
+            response.status_code,
+            "\n".join(["{}: {}".format(header, response.headers[header])
+                       for header in response.headers]))
         if payload:
-            logger.debug("Response Body: %s", pprint.pformat(response.content))
+            msg += "\n{}".format(pprint.pformat(response.content))
```
begoldsm: be careful with this, since download (op=OPEN) has a payload that can be quite large. I would recommend protecting yourself here and do one of the following:
1. ignore the payload specifically for the OPEN op when it has a 200 response and write out that the output was ignored due to the method called
2. Have a blanket truncation rule where the payload is logged up to some amount of data (100KB, maybe?) with a message after indicating it was truncated.

jbcrail: I went with option 2 (limiting to 64K, defined by MAX_CONTENT_LENGTH).

jbcrail:
I've added output from the CLI when listing a non-existent file:
```
$ python azure/datalake/store/cli.py ls bar
ERROR:azure.datalake.store.lib:Exception FileNotFoundError('bar',)
404
Cache-Control: no-cache
Pragma: no-cache
Expires: -1
x-ms-request-id: de80377b-b5c4-44b3-a261-c70f1e84dcfd
Server-Perf: [de80377b-b5c4-44b3-a261-c70f1e84dcfd][ AuthTime::939.555812908222::PostAuthTime::219.814150129643 ][S-HdfsListStatus :: 00:00:006 ms]%0a[LISTSTATUS :: 00:00:006 ms]%0a
Status: 0x8309000A
Date: Thu, 06 Oct 2016 02:11:34 GMT
Content-Length: 219

{"RemoteException":{"exception":"FileNotFoundException","message":"Folder does not exist: /bar [de80377b-b5c4-44b3-a261-c70f1e84dcfd][2016-10-05T19:11:34.9419677-07:00]","javaClassName":"java.io.FileNotFoundException"}}
Traceback (most recent call last):
  File "azure/datalake/store/cli.py", line 474, in <module>
    AzureDataLakeFSCommand(fs).onecmd(' '.join(sys.argv[1:]))
  File "/Users/jcrail/anaconda2/envs/test/lib/python2.7/cmd.py", line 221, in onecmd
    return func(arg)
  File "azure/datalake/store/cli.py", line 273, in do_ls
    for item in self._fs.ls(d, detail=args.detail):
  File "/Users/jcrail/src/azure-data-lake-store-python/azure/datalake/store/core.py", line 116, in ls
    files = self._ls(path)
  File "/Users/jcrail/src/azure-data-lake-store-python/azure/datalake/store/core.py", line 107, in _ls
    out = self.azure.call('LISTSTATUS', path.as_posix())
  File "/Users/jcrail/src/azure-data-lake-store-python/azure/datalake/store/lib.py", line 292, in call
    self.log_response_and_raise(r, FileNotFoundError(path))
  File "/Users/jcrail/src/azure-data-lake-store-python/azure/datalake/store/lib.py", line 251, in log_response_and_raise
    raise exception
azure.datalake.store.exceptions.FileNotFoundError: bar
```

begoldsm:
Looks good, thanks @jbcrail. In the future we can probably reduce the number of headers returned to reduce the size of the error, but this is returning all the information the customer will need.

Issue #96, opened 2016-10-06
Auth with user name and password does not work


avaranovich:
`token = lib.auth(tenant_id, username, password)`

gives

`adal.adal_error.AdalError: Get Token request returned http error: 400 and server response: {"error":"invalid_request","error_description":"AADSTS90014: The request body must contain the following parameter: 'client_id'.\r\nTrace ID: 899e028b-469f-4dba-ab97-518e5a3e42e5\r\nCorrelation ID: 24bd2ce3-b084-40ab-9e8f-46f5a54ac501\r\nTimestamp: 2016-10-06 09:21:19Z","error_codes":[90014],"timestamp":"2016-10-06 09:21:19Z","trace_id":"899e028b-469f-4dba-ab97-518e5a3e42e5","correlation_id":"24bd2ce3-b084-40ab-9e8f-46f5a54ac501"}`

and according to this line https://github.com/Azure/azure-data-lake-store-python/blob/dev/azure/datalake/store/lib.py#L106

it cannot work just with username and password.

begoldsm:
@jbcrail @martindurant this appears to be an issue with how we are defaulting the client ID. I am going to make a small change to properly default the client ID, which should resolve this issue.

@avaranovich to work around this please pass in the following client_id during authentication: 1950a258-227b-4e31-a9cf-717495945fc2

begoldsm:
@avaranovich this should now be resolved with the latest build.

PR #97, opened 2016-10-06
Update to use the default client_id (issue #96)

begoldsm:
This will use the default client id if the following conditions are met:
1. The user does not add one to an environment variable
2. The user does not pass one in during lib.auth().


begoldsm:
Since it didn't link, this should resolve issue #96

PR #99, opened 2016-10-13
Fix headers to ensure unique IDs

begoldsm:
by modifying the global self.head property within the Call function, there is no guarantee that it is not modified again by a different thread before the current thread finishes logging and execution. assigning to a local variable and using that should alleviate this issue.


begoldsm:
@jbcrail can you verify that this makes sense and also do a quick sanity pass to make sure there aren't other places where concurrency could clobber the expected values of global members?

PR #100, opened 2016-10-18
PR: Catch connection errors to short-circuit retries

jbcrail:
This was an error noticed in the debug logs for #85.


jbcrail:
Closing since the fix was too aggressive.

PR #102, opened 2016-10-19
PR: Miscellaneous fixes

jbcrail:
When testing bad offset errors, I find a few minor issues:

- explicitly copy headers dictionary (by default Python copies reference)
- clean up exception messages (formatted messages had raw `%s` in string)
- removed one level of indentation from `flush()`
- use `writable()` check instead of comparing against hard-coded set of file modes

### DIFF
--- a/azure/datalake/store/core.py
+++ b/azure/datalake/store/core.py
@@ -617,69 +617,73 @@ class AzureDLFile(object):
         If force=True, flushes all data in the buffer, even if it doesn't end
         with a delimiter; appropriate when closing the file.
         """
-        if self.mode in {'wb', 'ab'} and not self.closed:
-            if self.buffer.tell() == 0:
-                if force and self.first_write:
+        if not self.writable() and self.closed:
+            return
+
+        if self.buffer.tell() == 0:
+            if force and self.first_write:
+                _put_data(self.azure.azure,
+                          'CREATE',
+                          path=self.path.as_posix(),
+                          data=None,
+                          overwrite='true',
+                          write='true')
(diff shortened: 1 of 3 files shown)


azure/datalake/store/core.py
```diff
@@ -617,69 +617,73 @@ def flush(self, force=False):
         If force=True, flushes all data in the buffer, even if it doesn't end
         with a delimiter; appropriate when closing the file.
         """
-        if self.mode in {'wb', 'ab'} and not self.closed:
-            if self.buffer.tell() == 0:
-                if force and self.first_write:
+        if not self.writable() and self.closed:
```
begoldsm: should be an or now, right?

azure/datalake/store/lib.py
```diff
@@ -282,35 +282,35 @@ def call(self, op, path='', **kwargs):
         func = getattr(requests, method)
         url = self.url + path
         try:
-            headers = self.head
+            headers = self.head.copy()
```
begoldsm: Thanks for catching this @jbcrail! Is there any where else where we should take a look to make sure we aren't modifying static members during concurrent calls?

azure/datalake/store/core.py
jbcrail: Good catch. This was a copy error from the VM to my local machine.

azure/datalake/store/lib.py
jbcrail: I couldn't find other shared state or static members that would be impacted by concurrent calls.

PR #105, opened 2016-10-21
PR: Add optional path to df

jbcrail:
This will allow us to get the content summary for a particular path. This prevents the test recordings from taking too long if the account has a large amount of data.

### DIFF
--- a/azure/datalake/store/core.py
+++ b/azure/datalake/store/core.py
@@ -168,10 +168,10 @@ class AzureDLFileSystem(object):
         else:
             return {p['name']: p['length'] for p in files}
 
-    def df(self):
-        """ Resource summary
-        """
-        return self.azure.call('GETCONTENTSUMMARY')['ContentSummary']
+    def df(self, path):
+        """ Resource summary of path """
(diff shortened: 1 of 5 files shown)


azure/datalake/store/core.py
```diff
@@ -168,10 +168,10 @@ def du(self, path, total=False, deep=False):
...
             return {p['name']: p['length'] for p in files}
 
-    def df(self):
-        """ Resource summary
-        """
-        return self.azure.call('GETCONTENTSUMMARY')['ContentSummary']
+    def df(self, path):
+        """ Resource summary of path """
+        path = AzureDLPath(path).trim()
+        return self.azure.call('GETCONTENTSUMMARY', path.as_posix())['ContentSummary']
```
begoldsm: Do we have any other calls that don't take in path? I just want to make sure that we take in path everywhere that it is supported (which should be for all the calls, I think).

jbcrail: It looks like all `core.py` methods except `ls()` require a path. The optional path was meant for some CLI commands, like `df` (also included already are `ls` and `du`).

PR #113, opened 2016-11-10
PR: Thread-local sessions

### LINKED ISSUE #112: Parallel upload/download opening too many TCP/IP connections
The current implementation does not submit web requests through a session, which results in a new session being created and destroyed each time a request is sent. This results in a lot of time being spent in getaddrinfo calls when that information can be re-used. The solution to this is to execute requests through a session like in the below example:
```
self.global_session = requests.Session()
self.global_session.put(<url>, params=params, headers=headers, data=data, stream=stream)
```

note, though, that Session is not guaranteed to be thread safe, so session awareness will need to be passed down through each thread, resulting in one session per thread.

jbcrail:
In addition to thread-local sessions, I also improved the benchmarking CLI so that we could toggle log messages, verification, checksums, or iterations. It also outputs benchmarking statistics.

Fixes #112

### DIFF
--- a/azure/datalake/store/lib.py
+++ b/azure/datalake/store/lib.py
@@ -112,6 +112,13 @@ def auth(tenant_id=default_tenant, username=default_username,
                 'time': time.time(), 'tenant': tenant_id, 'client': client_id})
     return out
 
+
+def init_session():
+    s = requests.Session()
+    s.mount('https://', requests.adapters.HTTPAdapter(pool_connections=64, pool_maxsize=64))
+    return s
+
+
 class DatalakeRESTInterface:
     """ Call factory for webHDFS endpoints on ADLS
 
@@ -150,6 +157,7 @@ class DatalakeRESTInterface:
     def __init__(self, store_name=default_store, token=None,
                  url_suffix=default_suffix, **kwargs):
         url_suffix = url_suffix or "azuredatalakestore.net"
+        self.session = init_session()
         if token is None:
             token = auth(**kwargs)
         self.token = token
@@ -163,6 +171,7 @@ class DatalakeRESTInterface:
 
     def _check_token(self):
         if time.time() - self.token['time'] > self.token['expiresIn'] - 100:
+            self.session = init_session()
             self.token = refresh_token(self.token)
             self.head = {'Authorization': 'Bearer ' + self.token['access']}
 
@@ -238,7 +247,7 @@ class DatalakeRESTInterface:
                              keys - allowed)
         params = {'OP': op}
         params.update(kwargs)
-        func = getattr(requests, method)
+        func = getattr(self.session, method)
         url = self.url + path
         try:
             headers = self.head.copy()
--- a/azure/datalake/store/lib.py
+++ b/azure/datalake/store/lib.py
@@ -13,7 +13,6 @@ Specific interfaces to the Data-lake Store filesystem layer and authentication c
 """
 
 # standard imports
-import json
 import logging
 import os
 import threading
--- a/azure/datalake/store/lib.py
+++ b/azure/datalake/store/lib.py
@@ -293,6 +293,11 @@ class DatalakeRESTInterface:
             return out
         return r
 
+    def __getstate__(self):
+        dic2 = self.__dict__.copy()
+        dic2.pop('local', None)
+        return dic2
+
 """
 Not yet implemented (or not applicable)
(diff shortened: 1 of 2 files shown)


begoldsm:
@brettcannon @zooba can you guys take a look at this as well and let us know your thoughts? @jbcrail thank you for making it thread local. I will apply this patch to my branch and run through the benchmarks right now.

azure/datalake/store/lib.py
```diff
@@ -161,10 +164,23 @@ def __init__(self, store_name=default_store, token=None,
             __name__,
             __version__)
 
+    @property
+    def session(self):
```
brettcannon: This function could be coded as:
```python
try:
    s = self.local.session
except AttributeError:
    adapter = requests.adapters.HTTPAdapter(
             pool_connections=MAX_POOL_CONNECTIONS,
             pool_maxsize=MAX_POOL_CONNECTIONS)
    s = requests.Session()
    s.mount(self.url, adapter)
    self.local.session = s
return s
```

azure/datalake/store/lib.py
```diff
@@ -161,10 +164,23 @@ def __init__(self, store_name=default_store, token=None,
...
+            s = requests.Session()
+            s.mount(self.url, adapter)
+            setattr(self.local, 'session', s)
+        return s
+
     def _check_token(self):
         if time.time() - self.token['time'] > self.token['expiresIn'] - 100:
             self.token = refresh_token(self.token)
             self.head = {'Authorization': 'Bearer ' + self.token['access']}
+            setattr(self.local, 'session', None)
```
brettcannon: `self.local.session = None`

azure/datalake/store/lib.py
```diff
@@ -276,6 +292,11 @@ def call(self, op, path='', **kwargs):
             return out
         return r
 
+    def __getstate__(self):
+        dic2 = self.__dict__.copy()
```
brettcannon: There's only the one dict, so best to call it `dict_` or `state`.

azure/datalake/store/lib.py
jbcrail: Thanks, @brettcannon! Simpler is better.

jbcrail:
Performance benchmarks for Windows (10 runs; elapsed time in seconds):
```
benchmark   min mean sd median max
down-1-50gb 189 199  6  198    208
down-50-1gb 125 166  29 155    226
up-1-50gb   134 173  22 171    216
up-50-1gb   169 257  97 237    519
```

A few runs failed due to ThrottledException errors.

begoldsm:
@jbcrail thank you for the run details! Last night there were some stability issues in Azure that caused the throttling you saw. I am doing a sanity run on Ubuntu right now and will post those results as well.

begoldsm:
@jbcrail there appears to be an issue with reporting retries, although I am not sure this is related to the current PR. There is an attribute error. I thought that this was fixed previously, so I think this might be a regression. Can you include a fix for this in the current PR, please?

[errored] file D:\Data\100GB\2.dat -> \begoldsmpytest\100GB\2.dat, chunk \begoldsmpytest\100GB\2.dat.segments.1bd35714eb
01432f9d52f847e4752873\2.dat_0 0: RuntimeError('Max number of ADL retries exceeded: exception AttributeError("\'NoneType
\' object has no attribute \'put\'",)',)

azure/datalake/store/lib.py
```diff
@@ -238,7 +255,7 @@ def call(self, op, path='', **kwargs):
                              keys - allowed)
         params = {'OP': op}
         params.update(kwargs)
-        func = getattr(requests, method)
+        func = getattr(self.session, method)
```
begoldsm: sometimes this is appears to be resulting in a "NoneType" during retry of operations based on the error message:
RuntimeError('Max number of ADL retries exceeded: exception AttributeError(**"\'NoneType\' object has no attribute \'put\'"**,)',)

jbcrail: I think I found the source. Initial runs show the error doesn't reappear.

jbcrail:
Performance benchmarks for Linux (10 runs; elapsed time in seconds):
```
benchmark   min mean sd  median max
down-1-50gb 123 129  4   128    136
down-50-1gb 128 130  2   130    133
up-1-50gb   117 14   26  134    216
up-50-1gb   125 220  120 138    441
```

begoldsm:
@jbcrail changes look good, thank you!

PR #130, opened 2017-01-24
Merge dev into master.

begoldsm:
We need to publish a new version with a fix for overwrite.


begoldsm:
Closed to ensure that all merges into master are properly handled from dev.

PR #156, opened 2017-04-26
Updates to support throttling management

begoldsm:
These changes enable support for the potential throttling of an upload or download from the server. We have enabled back-off retry logic by default now. In addition, downloads that are being throttled can result in a smaller number of bytes being returned than was requested. This needs to be checked for and the download request needs to be retried until all the expected bytes have been received.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,10 @@
 
 Release History
 ===============
+0.0.7 (2017-04-19)
+------------------
+* Update the build process to more efficiently handle multi-part namespaces for pip.
+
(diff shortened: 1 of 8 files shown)


brettcannon:
Sorry I didn't get to this, @begoldsm ; got back from vacation on Thursday and been digging myself out of the work item backlog ever since. ☹️

begoldsm:
@brettcannon no worries! @lmazuel signed off which gave me some confidence. If you have any other comments though please feel free to add them and I can incorporate them into a future PR!

brettcannon:
PR LGTM 😄

PR #157, opened 2017-05-10
Work around for issue with pyOpenSSL

begoldsm:
Issue:  https://github.com/pyca/pyopenssl/issues/625 causes a
perf issue in our SDK when pyOpenSSL is installed. 
This workaround allows us to ensure that we do not use pyOpenSSL when it is installed until the above mentioned issue is investigated and resolved.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,7 +4,7 @@ Release History
 ===============
 0.0.9 (2017-05-09)
 ------------------
-* Enforce basic SSL utilization to ensure performance due to github issue: https://github.com/pyca/pyopenssl/issues/625
+* Enforce basic SSL utilization to ensure performance due to `GitHub issue 625 <https://github.com/pyca/pyopenssl/issues/625>`
 
 0.0.8 (2017-04-26)
 ------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,10 @@
 
 Release History
 ===============
+0.0.9 (2017-05-09)
+------------------
+* Enforce basic SSL utilization to ensure performance due to github issue: https://github.com/pyca/pyopenssl/issues/625
+
 0.0.8 (2017-04-26)
 ------------------
 * Fix server-side throttling retry support. This is not a guarantee that if the server is throttling the upload (or download) it will eventually succeed, but there is now a back-off retry in place to make it more likely.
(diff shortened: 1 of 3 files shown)


HISTORY.rst
```diff
@@ -2,6 +2,10 @@
 
 Release History
 ===============
+0.0.9 (2017-05-09)
+------------------
+* Enforce basic SSL utilization to ensure performance due to github issue: https://github.com/pyca/pyopenssl/issues/625
```
brettcannon:
```restructuredtext
due to `GitHub issue 625 <https://github.com/pyca/pyopenssl/issues/625>`_
```

azure/datalake/store/lib.py
```diff
@@ -25,6 +25,14 @@
 import requests
 import requests.exceptions
 
+# this is required due to github issue, to ensure we don't lose perf from openPySSL: https://github.com/pyca/pyopenssl/issues/625
+try:
+    from requests.packages.urllib3.contrib.pyopenssl import extract_from_urllib3 as enforce_no_py_open_ssl
+    enforce_no_py_open_ssl()
```
brettcannon: I would move this into an `else` clause in case it triggers an `ImportError` which is an actual error and not something to ignore.

azure/datalake/store/lib.py
```diff
@@ -25,6 +25,14 @@
...
+except ImportError:
+    # if OpenSSL is unavailable an import error will occur and that is ok.
+    pass
```
brettcannon: Would it be worth logging this?

begoldsm: We don't really care, since we just want to ensure that OpenSSL isn't used. If it doesn't exist, then we know that it isn't going to be used.

azure/datalake/store/lib.py
begoldsm: I am not quite sure I understand. You mean that I call enforce_no_py_open_ssl() in an `else` somewhere, or in an `if`?

brettcannon:
```python
try:
    from requests.packages.urllib3.contrib.pyopenssl import extract_from_urllib3 as enforce_no_py_open_ssl
except ImportError:
    pass
else:
    enforce_no_py_open_ssl()
```

PR #161, opened 2017-05-26
Fix local caching, performance improvements, bug fixes

begoldsm:
* Allow users to explicitly use or invalidate the internal, local cache of the filesystem that is built up from previous `ls` calls. It is now set to always call the service instead of the cache by default.
* Update folder upload/download to properly throw early in the event that the destination files exist and overwrite was not specified. NOTE: target folder existence (or sub folder existence) does not automatically cause failure. Only leaf node existence will result in failure.
* Fix a bug that caused file not found errors when attempting to get information about the root folder.


README.rst
```diff
@@ -31,6 +31,7 @@ To play with the code, here is a starting point:
     # typical operations
     adl.ls('')
     adl.ls('tmp/', detail=True)
+	adl.ls('tmp/', detail=True, invalidate_cache=True)
```
zooba: Losing the indent here will probably break your readme

azure/datalake/store/core.py
```diff
@@ -107,73 +107,103 @@ def open(self, path, mode='rb', blocksize=2**25, delimiter=None):
...
                            delimiter=delimiter)
 
-    def _ls(self, path):
+    def _ls(self, path, invalidate_cache=True):
         """ List files at given path """
         path = AzureDLPath(path).trim()
         key = path.as_posix()
+        
+        if invalidate_cache:
+            self.invalidate_cache(key)
```
zooba: You're going from never doing this to always doing it (by default)? Is that deliberate?

azure/datalake/store/core.py
```diff
@@ -107,73 +107,103 @@ def open(self, path, mode='rb', blocksize=2**25, delimiter=None):
...
         if not files:
-            inf = self.info(path)
+            # in this case we just invalidated the cache (if it was true), so no need to do it again
+            inf = self.info(path, invalidate_cache=False)
             if inf['type'] == 'DIRECTORY':
-                return []
+                if detail:
+                    return inf
+                else:
+                    return []
```
zooba: Reads more nicely as `return inf if detail else []`

tests/test_multithread.py
```diff
@@ -146,6 +146,7 @@ def test_download_many(tempdir, azure):
 
 @my_vcr.use_cassette
 def test_download_glob(tempdir, azure):
+    time.sleep(10)
```
zooba: Still necessary? Or left over from testing stuff

azure/datalake/store/core.py
begoldsm: Yes, because our service is a distributed filesystem being touched by many users across many different machines, it doesn't make sense to default to using the cached copy unless the user explicitly wants to and understands the risks of "staleness".

azure/datalake/store/core.py
begoldsm: That does read more nicely. Fixt!

tests/test_multithread.py
begoldsm: Good catch! that is left over from me needing to attach a debugger :)

azure/datalake/store/core.py
```diff
@@ -107,73 +107,101 @@ def open(self, path, mode='rb', blocksize=2**25, delimiter=None):
...
+
         if key not in self.dirs:
             out = self.azure.call('LISTSTATUS', key)
             self.dirs[key] = out['FileStatuses']['FileStatus']
             for f in self.dirs[key]:
                 f['name'] = (path / f['pathSuffix']).as_posix()
         return self.dirs[key]
 
-    def ls(self, path="", detail=False):
+    def ls(self, path="", detail=False, invalidate_cache=True):
```
brettcannon: Can't wait until you get to drop Python 2.7 support as this would have made a great keyword-only argument.

azure/datalake/store/core.py
```diff
@@ -107,73 +107,101 @@ def open(self, path, mode='rb', blocksize=2**25, delimiter=None):
...
         path = AzureDLPath(path).trim()
-        root = path.parent
         path_as_posix = path.as_posix()
-        for f in self._ls(root):
+        root = path.parent
+        root_as_posix = root.as_posix()
+        
+        # in the case of getting info about the root itself or if the cache won't be hit
+        # simply return the result of a GETFILESTATUS from the service
+        if invalidate_cache or path_as_posix == '/' or path_as_posix == '.':
```
brettcannon: You can also do `if invalidate_cache or path_as_posix in {'/', '.'}:`

azure/datalake/store/core.py
```diff
@@ -107,73 +107,101 @@ def open(self, path, mode='rb', blocksize=2**25, delimiter=None):
...
+            to_return  = self.azure.call('GETFILESTATUS', path_as_posix, expected_error_code=expected_error_code)['FileStatus']
+            to_return['name'] = path_as_posix
+            
+            # add the key/value pair back to the cache so long as it isn't the root
+            if path_as_posix != '/' and path_as_posix != '.':
```
brettcannon: `if path_as_posix not in {'/', '.'}:`

azure/datalake/store/core.py
```diff
@@ -107,73 +107,101 @@ def open(self, path, mode='rb', blocksize=2**25, delimiter=None):
...
+                if root_as_posix not in self.dirs:
+                    self.dirs[root_as_posix] = [to_return]
+                else:
+                    found = False
+                    for f in self.dirs[root_as_posix]:
+                        if f['name'] == path_as_posix:
+                            found = True
+                            f = to_return
+                    if not found:
+                        self.dirs[root_as_posix].append(to_return)
```
brettcannon: If you just need the first `f` to work instead of the last:
```python
for f in self.dirs[root_as_posix]:
    if f['name'] == path_as_posix:
        f = to_return
        break
else:
    self.dirs[root_as_posix].append(to_return)
```
And if you do need the last one, then simply pass `self.dirs[root_as_prefix]` through `reversed()`.

azure/datalake/store/core.py
```diff
@@ -635,7 +665,14 @@ def __init__(self, azure, path, mode='rb', blocksize=2**25,
...
+        
+        # cannot create a new file object out of a directory
+        if exists and self.info()['type'] == 'DIRECTORY':
+            raise IOError('path: {} is a directory, not a file, and cannot be opened for reading or writing'.format(path))
+        # always invalidate the cache when checking for existence of a file
+        # that may be created or written to (for the first time).
+        if mode == 'ab' and exists:    
             self.loc = self.info()['length']
             self.first_write = False
         if mode == 'rb':
```
brettcannon: `elif` since the previous branch will only match if `mode == 'ab'`.

azure/datalake/store/core.py
begoldsm: good catch, I missed the break previously. We are looking for the first f, since the name property is unique within a directory (we are basically going through a list of files/folders in a specific directory and looking for the one that matches).

PR #164, opened 2017-06-02
Feature request: put files being downloaded into "in progress" state

begoldsm:
This change puts files that are being downloaded but have not yet finished into an "in progress" state, finally renaming them at the end. This ensures that if there is a failure, the user will not see the files as "valid".

### DIFF
--- a/azure/datalake/store/transfer.py
+++ b/azure/datalake/store/transfer.py
@@ -427,7 +427,7 @@ class ADLTransferClient(object):
                         overwrite=self._parent._overwrite)
                     self._ffutures[merge_future] = parent
                 else:
-                    if dst.endswith('.inprogress'):
+                    if not self._chunked and str(dst).endswith('.inprogress'):
                         logger.debug("Renaming file to remove .inprogress: %s", self._fstates[parent])
                         self._fstates[parent] = 'merging'    
                         self._rename_file(dst, dst.replace('.inprogress',''), overwrite=self._parent._overwrite)
--- a/azure/datalake/store/transfer.py
+++ b/azure/datalake/store/transfer.py
@@ -404,12 +404,13 @@ class ADLTransferClient(object):
                                                       key=operator.itemgetter(1))], 
                         overwrite=self._parent._overwrite)
                     self._ffutures[merge_future] = parent
-                elif self._merge and not self._chunked and dst.endswith('.inprogress'):
-                    logger.debug("Renaming file to remove .inprogress: %s", self._fstates[parent])
-                    self._fstates[parent] = 'merging'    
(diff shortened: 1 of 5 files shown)


azure/datalake/store/transfer.py
```diff
@@ -361,6 +362,27 @@ def progress(self):
...
+            # we do a final check to make sure someone didn't create the destination file while download was occuring
+            # if the user did not specify overwrite.
+            if os.path.isfile(dst):
+                if not overwrite:
+                    raise FileExistsError(dst)
+                os.remove(dst)
+            os.rename(src, dst)
+        except Exception as e:
+            exception = repr(e)
+            logger.error('Rename failed for source file: %s; %s', src, exception)
```
brettcannon: Use `"Rename failed for source file %r: %r"` and you can avoid the `exception` variable entirely.

azure/datalake/store/transfer.py
```diff
@@ -361,6 +362,27 @@ def progress(self):
...
+            raise e
+    
+        logger.debug('Renamed %s to %s', src, dst)
```
brettcannon: Use `%r` in case their is unforeseen issues with some wonky whitespace used in the file path.

azure/datalake/store/transfer.py
```diff
@@ -361,6 +362,27 @@ def progress(self):
...
+        return None
```
brettcannon: Python always returns `None` if no return statement is specified, so only bother with explicitly return `None` when you **really** want to clearly return `None, else just use a bare `return` or just leave it out if it's the last statement of your function.

azure/datalake/store/transfer.py
```diff
@@ -405,6 +427,12 @@ def _update(self, future):
                         overwrite=self._parent._overwrite)
                     self._ffutures[merge_future] = parent
                 else:
+                    if not self._chunked and str(dst).endswith('.inprogress'):
```
brettcannon: Combine with `else` to make an `elif`.

azure/datalake/store/transfer.py
begoldsm: Thanks @brettcannon for the idea. Although I am a little confused how I can get rid of the exception variable, does %r: %r autoamtically give me the exception for the second %r here?

begoldsm: ah never  mind, I get it!

azure/datalake/store/transfer.py
begoldsm: This actually needs to be separate, since there is something that needs to always happen in the else (marking the file as finished) but there is something that only needs to happen in the else in the event that there is a file with the suffix name  .inprogress. If it doesn't end with that suffix, then we just update the transfer state to 'finished'.

Issue #165, opened 2017-06-15
Automatically refresh service principal access token


matt1883:
The built-in authentication logic accepts service principal credentials to fetch an access token once. When this access token expires, the authentication logic does not refresh the access token.

The access token should be refreshed on-demand, if the existing token has expired.

This allows the user of the SDK to not worry about manually refreshing the token in the case of service principal auth, much like the existing support for refreshing user tokens.

Note: Special consideration may be needed re: how the service principal's credentials are stored for re-use when fetching a new token. Other Azure SDKs for Python (e.g., Azure SDK's auth library) may provide insight here.

begoldsm:
This is resolved and published by PR #167

PR #167, opened 2017-06-29
Add ServicePrincipal token refresh support

begoldsm:
This change addresses issue: https://github.com/Azure/azure-data-lake-store-python/issues/165 and adds a quality of life improvement to automatically refresh service principal credentials when they expire within the client, removing that responsibility from the caller of the SDK.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,10 @@
 
 Release History
 ===============
+0.0.13 (2017-06-28)
+-------------------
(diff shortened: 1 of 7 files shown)


brettcannon:
Sorry for not getting to this, @begoldsm , but this came in during crunch time for PyData Seattle. Do you still want a review?

begoldsm:
@brettcannon I think it should be ok for now. @lmazuel  gave me a quick offline review and I think we are good! Thanks though!

Issue #174, opened 2017-07-13
Enable Data Lake Store progress controller callback


clehene:
### Description
This should report progress on transferred chunks (similar to blob store transfer)
```
Alive[#########################                                       ]  40.0881%
Finished[#############################################################]  100.0000%
```

begoldsm:
@clehene  This is a an extension of enhancement: #142 for the CLI.

Currently progress support will show up if --verbose is specified in the CLI or verbose=True is passed in to the SDK directly, however it drastically reduces overall performance for folder upload/download when there are a large number of files, due to the single threaded nature of Python and the constant context switching between the callbacks and the actual upload/download.

That being said this is something the team is aware of and is part of the current planning discussion for the upcoming milestone. @matt1883 will be able to share more details once we have them.

begoldsm:
Just saw that you also submitted a PR @clehene, thank you! The PR is: #175

clehene:
@begoldsm it looks like the `progress`, `status` functions are pretty heavy as they will iterate across all files / chunks with every operation. The implementation in the PR makes use of the core progress controller and only updates transferred bytes, so IMHO shouldn't affect performance dramatically. Is there a benchmark test implemented so we could compare?

begoldsm:
@clehene that is great to know! We do have a benchmark test although, unfortunately, it is not entirely self contained, as it assumes the existence of certain files with specific names. it exists in samples/benchmarks.py, and could easily be updated, I think, to create the files that it uses to test on execution instead of relying on them to be pre-existing. If you are up for making some additional improvements I won't stop you! 😄

begoldsm:
#175 enables this.

PR #175, opened 2017-07-13
Enable Data Lake Store progress controller callback #174

clehene:
---

### General Guidelines

- [] Links to associated bugs, if any, are in the description.


azure/datalake/store/transfer.py
```diff
@@ -240,9 +245,14 @@ def __init__(self, adlfs, transfer, merge=None, nthreads=None,
         self._chunked = chunked
         self._unique_temporary = unique_temporary
         self._unique_str = uuid.uuid4().hex
+        self._progress_callback=progress_callback
+        self._progress_lock = threading.Lock()
```
begoldsm: I am concerned that this will result in a big performance hit for folders containing lots of small items (such as a folder with 100,000 1MB files). I like the idea, but we will probably need logic in the CLI to either disable progress by default or be smart enough to disable it in situations with known performance implications.

begoldsm:
@clehene can you please take a look at the failing test? The code currently uses pickle to save state of upload/download operations and it appears that you can't pickle lock objects. Please also add corresponding tests that validate the changes that you have made.

Thanks!

azure/datalake/store/transfer.py
clehene: Is there an existing benchmark for this that I could test with?

IMHO the lock contention penalty should be small compared to the IO generated by transfer which will be a non-trivial HTTP operation so I'd expect this to have a small impact if any.

That said, `_update` is rather elaborated and has multiple operations updating shared state. On a first look it doesn't seem necessarily problematic but something like this:

```
 self._chunks[obj]['actual'] = nbytes
 self._chunks[obj]['exception'] = exception
```
won't execute atomically. 

Ultimately, if this locking doesn't affect performance, perhaps may be more appropriate to happen in `_update` rather than `_update_progress`.

azure/datalake/store/transfer.py
```diff
@@ -382,6 +393,17 @@ def _rename_file(self, src, dst, overwrite=False):
     
         logger.debug('Renamed %r to %r', src, dst)
 
+    def _update_progress(self, length):
+        if self._progress_callback is not None:
+            if self._progress_lock is not None:
```
brettcannon: You seem to unconditionally set `_progress_lock` so checking for its existence is unnecessary.

azure/datalake/store/transfer.py
```diff
@@ -382,6 +393,17 @@ def _rename_file(self, src, dst, overwrite=False):
...
+                with self._progress_lock:
+                    self._progress_total_bytes += length
+                    total = self._progress_total_bytes
```
brettcannon: Why the local assignment and not just pass `self._progress_total_bytes` to the callback? Also, [assignment is atomic in CPython](https://docs.python.org/2/faq/library.html#what-kinds-of-global-value-mutation-are-thread-safe).

azure/datalake/store/transfer.py
brettcannon: If you're willing to trade some memory for lower lock-contention then you can heavily scale back the use of the lock. `list.append()` is atomic, so if you append to a list the total bytes sent then you don't need to lock. And calling `sum()` is also atomic so you could do `sum(sent_chunks)`. And if memory is an issue you could then do a cheap check for when the buffer gets to a certain size and then collapse it. IOW:
```python
self._send progress_buffer = []
...
self._send_progress_buffer.append(length)  # Atomic
if len(self._send_progress_buffer) > 1000:  # Choose some number you're comfortable with.
    with self._progress_lock:
        # In-place replacement to keep non-locking guarantees.
        self._send_progress_buffer[:] = [sum(self._send_progress_buffer)]
# Assignment in this order guarantees that total >= sent if total increases dynamically.
sent = sum(self._send_progress)
total = self._progress_total_bytes
self._progress_callback(sent, total)
```
This example will never lose data, guaranteed (Python unofficial memory model is, "order isn't guaranteed, but you won't lose data"). And if self._progress_total_bytes changes its total as time progresses you at worst might end up slightly out of step with bytes sent versus the total at the time of the send, but as long as the total is dealt with first you shouldn't ever end up with more bytes "sent" than there are in total.

clehene: @brettcannon thank you for the hints and details about the memory model. Being a casual Python user, this is helpful.

I thought about something similar as an alternative (I was thinking about keeping the progress as part of the existing chunk state. I'd be ok with either way, if this proves to have an impact on performance. If it doesn't I'd lean towards the simpler implementation, but glad to do whatever you think it's best.

Not quite familiar with the Python locking implementation but was assuming a lightweight approach initially and an uncontended lock may be cheap in this case. Even with a large number of small files, we're still limited by the actual number of threads that are performing IO operations in the order of milliseconds at least while an unconteded lock seems to be at least two orders of magnitude faster (0.1 us - https://mail.python.org/pipermail/python-ideas/2010-July/007644.html).

 I'd run some tests just to satisfy my curiosity.

clehene:
* Fixed tests (Added a Lock wrapper class that avoids serialization)
* Added a unit test to validate progress callback
* Updated README with instructions
* Ran performance tests on degenerate (worst case scenario) test with 10,000 files (see my other comment) that's showing locking doesn't affect performance. (details [here](https://github.com/Azure/azure-data-lake-store-python/pull/175#discussion_r127835327))
* Other fixes based on the previous comments.

Please re-review

begoldsm:
@clehene thank you so much this is a wonderful addition! I will review this today and I am so happy to see that performance does not regress (as you suspected! 😄 ).
@brettcannon would you like to take another look as well?

azure/datalake/store/transfer.py
begoldsm: @clehene this is wonderful and thank you so much for being so diligent in contributing to the repository, it is very appreciated!

README.rst
```diff
@@ -54,6 +54,36 @@ To play with the code, here is a starting point:
...
+When passed, this will keep track of transferred bytes and be called on each complete chunk.
+
+Here's an example using the Azure CLI progress controller as the `progress_callback`:
+
+.. code-block:: python
+
+    from cli.core.application import APPLICATION
+
+    def _update_progress(current, total):
+        hook = APPLICATION.get_progress_controller(True)
```
brettcannon: Use the keyword argument for the `True` argument (and basically any time you pass an argument where it's not self-documenting what the argument is meant for, e.g. what's `True`?).

azure/datalake/store/transfer.py
```diff
@@ -111,6 +111,35 @@ def __str__(self):
 Chunk = namedtuple('Chunk', 'name state offset expected actual exception')
 
 
+class NoSerLock(object):
```
brettcannon: Class name isn't very descriptive, e.g. what does the "Ser" part mean?

azure/datalake/store/transfer.py
```diff
@@ -111,6 +111,35 @@ def __str__(self):
...
+    """
+    Pickle friendly wrapper for threading.Lock (which is non-inheritable
+     so we wrap relevant functions)
```
brettcannon: Missing periods at the end of the sentences.

azure/datalake/store/transfer.py
```diff
@@ -111,6 +111,35 @@ def __str__(self):
...
+    The main part of this wrapper is that __getstate__ will return an empty
+    tuple so that the actual lock isn't serialized
```
brettcannon: If this class is only used in a single class then would it make more sense to define `__getstate__()` for that class instead?

azure/datalake/store/transfer.py
```diff
@@ -111,6 +111,35 @@ def __str__(self):
...
+    """
+
+    def __init__(self):
+        self._lock = threading.Lock()
+
+    def acquire(self):
```
brettcannon: You never seem to call these wrapper methods that re-expose the lock so you might as well just leave them out. And if I'm misreading the code and you actually do then it would be easier to just not hide `self._lock` and call the methods on the underlying lock directly.

Another option, though, is to define `__getattr__()` and delegate all missing attributes to the lock delegate so that all you're doing in this class is defining `__init__()`, `__setstate__()`, and `__getstate__()`.

tests/test_transfer.py
```diff
@@ -63,6 +63,29 @@ def transfer(adlfs, src, dst, offset, size, blocksize, buffersize, shutdown_even
                                                for chunk in f.chunks])
 
 
+def test_update_progress(azure):
+    """
+    Upload a 32 bytes file in chunks of 8 and test that the progress is incrementally
+        updated
```
brettcannon: Unneeded indent and missing a period.

tests/test_transfer.py
```diff
@@ -63,6 +63,29 @@ def transfer(adlfs, src, dst, offset, size, blocksize, buffersize, shutdown_even
...
+    """
+    calls = []
+
+    def recording_callback(progress, total):
+        calls.append((progress, total))
+
+    def transfer(adlfs, src, dst, offset, size, blocksize, buffersize, shutdown_event=None):
+        time.sleep(0.01)
```
brettcannon: What is the `sleep()` call for?

begoldsm:
@clehene Brett has some good feedback points. I will hold off on merging until you have a chance to address the questions/recommendations. Thanks!

tests/test_transfer.py
clehene: There's a bug in the way tasks are submitted and how state is handled on the callback #177
The sleep is a workaround used throughout the tests to avoid a race condition throughout the tests that do `client.run()`. 
Initially I assumed there's an issue with how pytests runs the tests and kept the sleep. However on a second look I found the bug.
Since the bug is not introduced by this code,  I'd keep this as is and rely on having the fix as a subsequent commit.

azure/datalake/store/transfer.py
clehene: nevermind, removed class

azure/datalake/store/transfer.py
clehene: Ignore. Fixed.  I somehow missed the `__getstate__` defined for the outer class. Thanks.

azure/datalake/store/transfer.py
clehene: Removed this.

tests/test_transfer.py
clehene: Made a PR for #177 so if that gets merged first I can remove the `sleep` here.

clehene:
@begoldsm @brettcannon please re-review. Also take a look at the other PR

begoldsm:
This is great. Thank you very much @clehene I am ok with merging this first, then taking a look at #177. I would rather not block this PR on that one. Thanks again!

PR #191, opened 2017-08-30
add sync flag, session id in rest calls during read/write

chdevala:
---

### General Guidelines

### DIFF
--- a/azure/datalake/store/core.py
+++ b/azure/datalake/store/core.py
@@ -5,6 +5,7 @@
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 # --------------------------------------------------------------------------
+from test.test_dis import outer
 
 """
 The main file-system class and functionality.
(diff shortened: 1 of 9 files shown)


azure/datalake/store/core.py
```diff
@@ -899,7 +913,10 @@ def flush(self, force=False):
                         path=self.path.as_posix(),
                         data=d2,
                         overwrite='true',
-                        write='true')
+                        write='true',
+                        syncFlag=syncFlag,
```
asikaria-msft: You cannot just pass syncFlag as-is from the caller. If the buffer (data[] array) has more than 4MB, then this method breaks it up into 4MB chunks. In that case, only the last chunk should send syncFlag=CLOSE; all others before that should send syncFlag=DATA.

azure/datalake/store/enums.py
```diff
@@ -12,3 +12,9 @@ class ExpiryOptionType(Enum):
     relative_to_now = "RelativeToNow"
     relative_to_creation_date = "RelativeToCreationDate"
     absolute = "Absolute"
+
+
+class SyncFlac(Enum):
```
asikaria-msft: Are you using this anywhere? Your actual usage above seems to use the strings directly.

PR #241, opened 2018-09-13
Add parameter to set a maximum number of files to download or upload

claudioscalzo:
### Description of the change
Added the `nfiles` parameter to the `ADLDownloader` and `ADLUploader` classes in order to set an upper bound to the files that are transferred.

This is simply done by cutting the list of files that is internally generated by the `glob` or the `walk` function; if the parameter is not set (default to `None`) the list won't be cropped (`any_list[:None]` is `any_list`).

### Further improvement
A further improvement could be to allow the user to directly provide a list instead of a path (so moving the `glob`,  `walk`, `ls` or *custom list creation* effort to the user side), resulting in a better flexibility.

### General Guidelines

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+0.0.32 (2018-09-12)
++++++++++++++++++++
+* Added parameter to set a maximum number of files to download or upload
+
 0.0.31 (2018-09-10)
 +++++++++++++++++++
 * Added support for batched ls
--- a/azure/datalake/store/__init__.py
+++ b/azure/datalake/store/__init__.py
@@ -6,7 +6,7 @@
 # license information.
 # --------------------------------------------------------------------------
 
-__version__ = "0.0.31"
+__version__ = "0.0.32"
 
 from .core import AzureDLFileSystem
 from .multithread import ADLDownloader
--- a/azure/datalake/store/multithread.py
+++ b/azure/datalake/store/multithread.py
@@ -99,6 +99,8 @@ class ADLDownloader(object):
         Callback for progress with signature function(current, total) where
         current is the number of bytes transfered so far, and total is the
         size of the blob, or None if the total size is unknown.
+    nfiles: int [None]
+        Maximum number of files to download. If None, downloads all files.
 
     See Also
     --------
@@ -106,7 +108,7 @@ class ADLDownloader(object):
     """
     def __init__(self, adlfs, rpath, lpath, nthreads=None, chunksize=2**28,
                  buffersize=2**22, blocksize=2**22, client=None, run=True,
-                 overwrite=False, verbose=False, progress_callback=None):
+                 overwrite=False, verbose=False, progress_callback=None, nfiles=None):
         
         # validate that the src exists and the current user has access to it
         # this only validates access to the top level folder. If there are files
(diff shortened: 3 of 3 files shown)


akharit:
@claudioscalzo Thanks for the contribution! 

I was wondering why you want this feature though. Can you present a use case where this would be useful?

Also, it would be great if you can add a test case that tests download with various values of n. 

The main issue I see here is how to include the number of directories in this metric. Do we create the whole directory structure, or just for the n files. Then there are issues with ordering etc. as LISTSTATUS does not provide any inherent ordering guarantees.

claudioscalzo:
@akharit Thanks for the answer!

This feature is not related to a particular use case and can be useful when the user is simply interested in a subset of the path and there is no need to start a download of the entire path or match the file names with wildcards. I see your concerns about the directory structure re-creation and about the LISTSTATUS lack of guarantees, and in those cases the parameter should be used when there is no need of both sorting and structure re-creation needs.

However, a much cleaner solution which I suggest (and it would be awesome to know if you agree on that) is to provide a way to the user to specify a list and not a path. In this way, the user would have the freedom to:
- create the list as he wants (using `ls`, `glob`, `walk` or any other different way)
- manipulate it (splitting it, elements removal, etc.)
- pass the job to the multi-threaded download/upload function

This feature is indeed related to a particular use case:
the number of files in the source folder can be huge, resulting in the user need to download just one batch of the folder content, process it, and then go ahead with the next batch. With the current implementation this seems to not be possible. With the proposed solution instead, the user could create the list of files, split it in the way he wants (like creating a list of lists), and call the download function on each list (batch).

From my point of view, this use case can happen very often in a big data environment, when folders can simply contain a number of files which is too big to be downloaded and processed in a one-shot fashion.

What do you think about this solution? Do you agree with me or do you see any drawbacks? In case of an affirmative answer I can transfer this discussion to an issue and close this PR. Thanks.

akharit:
adl = core.AzureDLFileSystem(token, store_name="")
    file_list = adl.ls('path')

    file_list = filter(filter_func, file_list)
    for files in file_list:
        ADLDownloader(adlfs=adl, rpath=files, lpath="./")

I am curious if there are any problems with the above approach. The user can get the file list and filter it as needed, and run the downloader on the files required. If more speed is needed(note that ADLDownloader is already multi-threaded), we can launch multiple instances of ADLDownloader as well. 

The problem with passing multiple files to ADLDownloader is mainly related to maintaining parity with other adls sdks. I'll have to check with other maintainers. 

However I don't see any particular advantage of that approach over the stub code above. Is there anything I am missing here?

claudioscalzo:
@akharit Thanks for the answer!

It will work, for sure, but the main concern is about the performances.

I see your proposal of running multiple instances of the ADLDownloader and in the past days, indeed, I already implemented some workarounds like yours, exploiting the parallelization of multiple ADLDownloaders dealing with single files (even if I have some related issues when dealing with folders with hundreds of thousands of files).

The purpose of my solution is to have a clean and reliable way to just pass the list and start the job with its multithreaded implementation, which avoids any potential error-prone code from the developer (who will try, for the sake of performances, to parallelize the instances in several ways like threads or cluster parallelization).

However, I understand your concern to maintain parity with other ADLS SDKs: it's an understandable reason and of course I leave the decision to you. Thanks.

akharit:
Closing as not the desired fix. 

I agree that ADLDownloader downloading multiple files is a better approach and I am looking into possibility of implementing that.

Issue #269, opened 2019-01-07
multithread.ADLUploader hangs indefinitely


djtrumbe:
### Description

multithread.ADLUploader hanging when uploading many (~40) files simultaneously.

Some background:

 - We are processing many json files into tabular, parquet files.
 - This processing is done using ~40 worker threads, each thread processing a single file before grabbing another file off of the queue.
 - Each thread, pulls down the json file locally, iterates over the lines in the file, converts them to python dictionaries, flattens those dictionaries and adds them to a list.
 - Once this thread has processed all rows into a list of dictionaries, it creates a pandas dataframe with the data, then writes the data to a temporary, local parquet file, then uploads that file to an ADL.

The problem comes when there are thousands of very small files, so the processing of the files occurs quickly and we are quickly trying to upload many files simultaneously.  The process hangs indefinitely, producing no error nor returning at all.

Here are also a few code snippets to give you some understanding of how this all works:

import multiprocessing as mp
...
pool = mp.Pool(threads)
...
parse_results = pool.map(func=multi_process_file, iterable=process_info, chunksize=1)
...
to_adl_parquet(df, adl_file, adls, log_tag)
...
def to_adl_parquet(df, adl_file, adl_connection, log_tag='', timestamp_fix=False):
    log_tag = log_tag and '{} -'.format(log_tag) or log_tag
    temp_file = '{}/{}_{}'.format(TEMP_DIRECTORY, uuid.uuid1(), os.path.basename(adl_file))
    log('{}   writing temp file: {}'.format(log_tag, temp_file))
    if timestamp_fix:
        df.to_parquet(temp_file, coerce_timestamps='us', allow_truncated_timestamps=True)
    else:
        df.to_parquet(temp_file)
    file_size = os.path.getsize(temp_file)
    log('{}   uploading file:    {}  ({} bytes)'.format(log_tag, adl_file, file_size))
    multithread.ADLUploader(adl_connection, lpath=temp_file, rpath=adl_file, nthreads=1, overwrite=True)  ## THIS IS THE LINE THAT HANGS INDEFINITELY
    os.remove(temp_file)

---

### Reproduction Steps
This problem is difficult to consistently reproduce.  However, the more small files we are processing, the more likely this is to happen.

### Environment summary

**SDK Version:** What version of the SDK  are you using? (pip show azure-datalake-store)

Name: azure-datalake-store
Version: 0.0.39
Summary: Azure Data Lake Store Filesystem Client Library for Python
[...]

Issue #265, opened 2018-12-17
Head (or similar) should support a timeout value


epa095:
### Description
We have an issue where at least one file on our datalake is broken. In the portal it has a size, but preview does not work, and "downloading" it results in an empty file. 

The problem is that when trying to download it (using `open`) with this library the result is that python just hangs with 100% CPU. Using info/ls/stat etc gives valid information, so there is no way of discovering its "broken-ness" from this. Using e.g. `head file_name 1` to read a single byte also results in python standing in an infinite loop with 100% CPU. It would be nice if `head`  accepted a timeout-parameter, then we could use it to detect when we were in this situation. Alternatively, do you have any other way of programmatically detect it?

### Reproduction Steps
** Enumerate the steps to reproduce the issue here:**

### Environment summary

**SDK Version:** What version of the SDK  are you using? (pip show azure-datalake-store)
Answer here:
azure-datalake-store==0.0.39

**Python Version:** What Python version are you using? Is it 64-bit or 32-bit?  
Answer here:
3.6.7 64bit

**OS Version:** What OS and version are you using?  
Answer here:
Linux

**Shell Type:** What shell are you using? (e.g. bash, cmd.exe, Bash on Windows)  
Answer here:
bash

akharit:
@epa095
Can you elaborate on what you mean when you say that "one file on our datalake is broken". In particular, it would be helpful if I can get the following info.

    adl = ADLS_INSTANCE
    print(adl.info(path_to_corrupted_file))

Can you also mention the adls account name, file path, and approximate time(with time zone) when you run the info and open operations? This will allow me to look at backend logs to figure out what exactly is happening.

Usually the backend server requests time out if necessary and propagate to the sdk, so explicit time outs aren't needed for functions. Here it seems like there is a bug in the sdk(which may be a result from bug in backend). From what I can figure out, the loop in read function i.e.

        while length > 0:
            self._read_blocksize()
            data_read = self.cache[self.loc - self.start:
                             min(self.loc - self.start + length, self.end - self.start)]
            out += data_read
            self.loc += len(data_read)
            length -= len(data_read)

is most probably not getting the length parameter reduced, and going into an infinite loop. This seems like the only pathway that will result in 100% Cpu usage. I can add a break on len(data_read) == 0 to fix, but I'll feel more confidant in the fix if I can get more info about the file and backend operations.

epa095:
[...]

PR #273, opened 2019-02-01
Release version 0.0.41

akharit:
---

### Description of the changes
* Remove GetContentSummary api call
* Move check_token() under retry block
* Expose timeout parameter for AdlDownloader and AdlUploader
* Raise an exception instead of silently break for zero length reads
* ~~Updated api-version to 2018-09-01~~
* ~~Updated all test recordings as api-version changed.~~
* Redid the tests so there are changes but tests still are based on the api-version 2018-05-01

Timeout may be helpful to debug #269 and also changed the fix for #265 to raise an exception in case of error.
### General Guidelines


azure/datalake/store/core.py
```diff
@@ -858,14 +884,18 @@ def read(self, length=-1):
...
+        flag = 0
         out = b""
         while length > 0:
             self._read_blocksize()
             data_read = self.cache[self.loc - self.start:
                              min(self.loc - self.start + length, self.end - self.start)]
             if not data_read:  # Check to catch possible server errors. Ideally shouldn't happen.
-                break
+                flag += 1
+            if flag >= 5:
```
rahuldutta90: inside the above if

PR #278, opened 2019-03-01
Release 0.0.43

akharit:
---

### Description of the change
Fix a bug with handling of glob.
There is a bug in the code where when we encounter len(rfiles)==1, we try to check whether the rpath contained a dir or a file. When rpath is a glob i.e contains a "*" or "?" this check fails as it tries to query the path as is. When we try to run info on paths like "/root/a/*/b.txt" it fails and raises an exception.

### General Guidelines


azure/datalake/store/multithread.py
```diff
@@ -195,26 +195,42 @@ def clear_saved():
...
         """
-        if "*" not in self.rpath:
-            rfiles = self.client._adlfs.walk(self.rpath, details=True, invalidate_cache=True)
-        else:
+
+        def is_glob_path(path):
+            path = AzureDLPath(path).trim()
+            prefix = path.globless_prefix
+            return not path == prefix
+
```
rahuldutta90: wont it be better to get is_glob_path(self.rpath) once and use that value twice

akharit: Agreed. Made changes

PR #283, opened 2019-04-26
Fix flush and open

akharit:
---

### Description of the change
flush was returning early if buffer is empty, even when called with force flag. Fixed.
Made sure that the delimiter test doesn't require on adls internal behavior.
opening a file required 3 separate calls to getFileStatus. Now it takes only one call.

### General Guidelines

will do when merge to master.

### DIFF
--- a/azure/datalake/store/core.py
+++ b/azure/datalake/store/core.py
@@ -956,14 +956,19 @@ class AzureDLFile(object):
 
         # always invalidate the cache when checking for existence of a file
         # that may be created or written to (for the first time).
-        exists = self.azure.exists(path, invalidate_cache=True)
+        try:
+            file_data = self.azure.info(path, invalidate_cache=True, expected_error_code=404)
+            exists = True
+        except FileNotFoundError:
+            exists = False
+        
 
         # cannot create a new file object out of a directory
-        if exists and self.info()['type'] == 'DIRECTORY':
+        if exists and file_data['type'] == 'DIRECTORY':
             raise IOError('path: {} is a directory, not a file, and cannot be opened for reading or writing'.format(path))
 
         if mode == 'ab' and exists:
-            self.loc = self.info()['length']
+            self.loc = file_data['length']
             self.first_write = False
         elif mode == 'rb':
             self.size = self.info()['length']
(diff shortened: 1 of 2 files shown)


azure/datalake/store/core.py
```diff
@@ -1188,7 +1192,6 @@ def flush(self, syncFlag='METADATA', force=False):
                     leaseid=self.leaseid,
                     filesessionid=self.filesessionid)
                 self.first_write = False
-            return
 
         self.buffer.seek(0)
         data = self.buffer.read()
```
rahuldutta90: can this be:
while len(data) > self.blocksize or force:
    if(len(data) < self.blocksize)
           force=false

azure/datalake/store/core.py
```diff
@@ -1229,9 +1232,8 @@ def flush(self, syncFlag='METADATA', force=False):
             logger.debug('Wrote %d bytes to %s' % (limit, self))
```
rahuldutta90: Maybe do this as a part of initialization of stream with no data:

if self.first_write:
                _put_data_with_retry(
                    self.azure.azure,
                    'CREATE',
                    path=self.path.as_posix(),
                    data=data[:limit],
                    overwrite='true',
                    write='true',
                    syncFlag=syncFlagLocal,
                    leaseid=self.leaseid,
                    filesessionid=self.filesessionid)
                self.first_write = False

tests/test_core.py
```diff
@@ -1137,15 +1137,21 @@ def test_array(azure):
```
rahuldutta90: do we have a test with 0 byte close and then check the length?

tests/test_core.py
```diff
@@ -1309,3 +1315,15 @@ def test_DatalakeBadOffsetExceptionRecovery(azure):
...
+
+
+def test_file_creation_open(azure):
+    with azure_teardown(azure):
+        if azure.exists(a):
+            azure.rm(a)
+        assert not azure.exists(a)
+        f = azure.open(a, "wb")
+        assert azure.exists(a)
+        f.close()
```
rahuldutta90: can you write something with data flag and then do close?

akharit: That is already done in previous tests. I was wrong on my earlier assumption that flush would not triggering the actual close for some files, as buffer is never empty except on first write operation, which was handled as a special case. 

I was artificially setting the buffer during my own code so that is why it was triggering. This wouldn't happen in a normal operation. 

This test confirms the change that the file is created at open, and not when the flush from close function is called.

PR #309, opened 2020-01-28
fix: Add positional arguments to NoRetryPolicy.should_retry

elvisace:
Add positional arguments to `NoRetryPolicy.should_retry`, so that an exception [here](https://github.com/Azure/azure-data-lake-store-python/blob/master/azure/datalake/store/lib.py#L429) will yield the correct behaviour.

Otherwise, the error will be `TypeError: should_retry() takes 1 positional argument but 4 were given` when `NoRetryPolicy` is used.

---

### General Guidelines

### DIFF
--- a/azure/datalake/store/retry.py
+++ b/azure/datalake/store/retry.py
@@ -21,7 +21,7 @@ logger = logging.getLogger(__name__)
 
 
 class RetryPolicy:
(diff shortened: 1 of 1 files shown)


akharit:
Thanks for fixing this!

PR #316, opened 2020-09-10
Fix error in retry when adal exception doesn't return response with json

akharit:
---

### Description of the change
Fix adal error when exception doesn't return correct response. ADAL was returning exception where e.error_response was probably an empty string. this happens when ADAL failed to parse the exception. Since this is a GET operation, we can still retry this.

### General Guidelines

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+0.0.50 (2020-09-10)
++++++++++++++++++++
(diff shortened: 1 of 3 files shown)


akharit:
Not sure why the auto check policy for travis is not working, but travis build link is https://travis-ci.org/github/Azure/azure-data-lake-store-python/builds/726073927

rahuldutta90 (approved):

Issue #329, opened 2023-03-27
Library depends on deprecated adal library


glaubitz:
The `adal` has been deprecated since 2021 and replaced by MSAL Python.

See: https://pypi.org/project/adal/

Thus, `azure-data-lake-store-python` should be switched to using MSAL.

akharit:
Resolved in 0.0.53

PR #330, opened 2023-04-11
Add MSAL support

akharit:
---

Updates MSAL
### Description of the change
Update MSAL
### General Guidelines

- [] The PR has supporting test coverage that confirm the expected behavior and protects against regressions, including necessary recordings.

Addresses #329 

Will add recordings separately as that just makes the pr harder to review.


azure/datalake/store/lib.py
```diff
@@ -136,60 +131,63 @@ def auth(tenant_id=None, username=None,
 
     if not client_secret:
         client_secret = os.environ.get('azure_client_secret', None)
+        contextPub = msal.PublicClientApplication(client_id=client_id, authority=authority+tenant_id)
```
rayluo, line 134: This PR is currently a 1:1 translation from ADAL to MSAL. It would work no worse than before, but it also does not utilize many good features from MSAL, such as token cache.

Ideally we shall reuse MSAL's `PublicClientApplication` and `ConfidentialClientApplication` (as mentioned in the first Q&A of [this recent conversation](https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/549)), but that would need a major architecture change to this repo. We discussed offline that this repo has a shelf life for less than a year, and [its replacement](https://learn.microsoft.com/en-us/samples/azure/azure-sdk-for-python/storage-datalake-samples/) is already available, so, we choose to use the current "1:1 translation" approach.

There is one low-hanging fruit we can grab. Define a module-wide in-memory dictionary, for MSAL to use it as an [http cache](https://msal-python.readthedocs.io/en/latest/#msal.PublicClientApplication.params.http_cache). This is required for MSAL's built-in http 429/5xx error retry to work across multiple `lib.auth(...)` runs.

```suggestion
        contextPub = msal.PublicClientApplication(client_id=client_id, authority=authority+tenant_id, http_cache=_http_cache)
```

azure/datalake/store/lib.py
```diff
@@ -136,60 +131,63 @@ def auth(tenant_id=None, username=None,
...
 
-    # You can explicitly authenticate with 2fa, or pass in nothing to the auth call
-    # and the user will be prompted to login interactively through a browser.
+    if tenant_id is None or client_id is None:
+        raise ValueError("tenant_id and client_id must be supplied for authentication")
 
-    @retry_decorator_for_auth(retry_policy=retry_policy)
+    contextClient = msal.ConfidentialClientApplication(client_id=client_id, authority=authority+tenant_id, client_credential=client_secret)
```
rayluo, line 139:
```suggestion
    else:
        contextClient = msal.ConfidentialClientApplication(client_id=client_id, authority=authority+tenant_id, client_credential=client_secret, http_cache=_http_cache)
```
Similar suggestion to the comment I put for `PublicClientApplication`. Also, this should be in an `else` clause, so that you can save overhead for constructing 2 MSAL classes but using only one.

rayluo (changes requested):

azure/datalake/store/multiprocessor.py
```diff
@@ -53,8 +53,8 @@ def log_listener_process(queue):
             queue.task_done()
             if record == END_QUEUE_SENTINEL:  # We send this as a sentinel to tell the listener to quit.
                 break
-            logger = logging.getLogger(record.name)
-            logger.handlers.clear()
+            logger = logging.getLogger("azure.datalake.store")
```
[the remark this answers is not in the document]
akharit: NA

azure/datalake/store/lib.py
```diff
@@ -136,60 +131,63 @@ def auth(tenant_id=None, username=None,
...
 
-    out.update({'access': out['accessToken'], 'resource': resource,
-                'refresh': out.get('refreshToken', False),
-                'time': time.time(), 'tenant': tenant_id, 'client': client_id})
+    if 'error' in out:
+        msg = "MSAL Error: "+out.get('error_description', "")
+        err = DatalakeRESTException(msg)
+        logger.log(logging.ERROR, msg)
+        raise err
+    print(out)
```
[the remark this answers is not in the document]
akharit: Removed

azure/datalake/store/lib.py
```diff
@@ -137,59 +133,60 @@ def auth(tenant_id=None, username=None,
...
+            out = contextPub.acquire_token_by_username_password(username=username, password=password, scopes=scopes)
         elif client_id and client_secret:
-            out = context.acquire_token_with_client_credentials(resource, client_id,
-                                                                client_secret)
+            out = contextClient.acquire_token_for_client(scopes=scopes)
             # for service principal, we store the secret in the credential object for use when refreshing.
             out.update({'secret': client_secret})
         else:
             raise ValueError("No authentication method found for credentials")
         return out
```
rayluo, line 155: While this new implementation will work in happy path, the separation of `PublicClientApplication` and `ConfidentialClientApplication` in MSAL happens to make the current implementation fragile. For example, if a user (accidentally?) configure `username` and `password` AND `client_secret`, the current implementation would run into a `contextPub` undefined exception.

Let's do some refactoring.

```python
scopes = kwargs.get('scopes', ["https://datalake.azure.net/.default"])
if client_secret:
    app = msal.ConfidentialClientApplication(
        client_id=client_id, authority=authority+tenant_id, client_credential=client_secret, http_cache=_http_cache)
    out = contextClient.acquire_token_for_client(scopes=scopes)
    # for service principal, we store the secret in the credential object for use when refreshing.
    out.update({'secret': client_secret})
else:
    app = msal.PublicClientApplication(client_id=client_id, authority=authority+tenant_id, http_cache=_http_cache)
    if require_2fa or (username is None and password is None):
        flow = app.initiate_device_flow(scopes=scopes)
        print(flow['message'])
        out = app.acquire_token_by_device_flow(flow)
    elif username and password:
        out = app.acquire_token_by_username_password(username=username, password=password, scopes=scopes)
    else:
        raise ValueError("No authentication method found for credentials")
```

akharit: This does change the order though as previously if both client secret and require_2fa were specified, we would go the 2fa route. 

I am pushing a diff with some code duplication but it preserves the previous conditional behavior.

rayluo: Setting both client_secret and require_2fa was not a valid setup in the first place. And now we have to become backward compatible with the old implementation detail which happened to run the 2fa code path?

That being said, you raised a good point, given that this is an older library that will retire within a year. Lesson learned. We can go with your new commit.

rayluo (approved):
Thanks for your effort and quick turnaround! Let's ship it!

PR #333, opened 2024-07-12
Use Azure Identity's TokenCredential for auth.

### LINKED ISSUE #332: Can't use Managed Identity and supply token for Auth
### Description
Can't use Managed Identity and supply token for Auth
---

### Reproduction Steps
Generate and supply token to base file-system class
```python
token_credential = get_token_credential()
access_token = token_credential.get_token("https://datalake.azure.net/.default")
adls_client = core.AzureDLFileSystem(access_token, store_name=adls_account_name)
```

Got error:
```
    adls_client = core.AzureDLFileSystem(access_token, store_name=adls_account_name)
  File "/usr/lib/python3.9/site-packages/azure/datalake/store/core.py", line 72, in __init__
    self.connect()
  File "/usr/lib/python3.9/site-packages/azure/datalake/store/core.py", line 90, in connect
    self.azure = DatalakeRESTInterface(token=self.token, req_timeout_s=self.per_call_timeout_seconds, **self.kwargs)
  File "/usr/lib/python3.9/site-packages/azure/datalake/store/lib.py", line 285, in __init__
    self._check_token()  # Retryable method. Will ensure that signed_session token is current when we set it on next line
  File "/usr/lib/python3.9/site-packages/azure/datalake/store/lib.py", line 313, in _check_token
    cur_session = self.token.signed_session()
AttributeError: 'AccessToken' object has no attribute 'signed_session'
```

### Environment summary

**SDK Version:** What version of the SDK  are you using? (pip show azure-datalake-store)
0.0.53

**Python Version:** What Python version are you using? Is it 64-bit or 32-bit?  
3.8

**OS Version:** What OS and version are you using?  
WSL+Ubuntu 22.04

**Shell Type:** What shell are you using? (e.g. bash, cmd.exe, Bash on Windows)

akharit:
Use Azure Identity's TokenCredential for auth.
Remove older python package support

Will fix #332 
---

### General Guidelines

- [ x] The PR has supporting test coverage that confirm the expected behavior and protects against regressions, including necessary recordings.

Not checking in the recordings. Fix needed wrt some fields like lease id, session id, token etc. that are causing the test run with recording to fail.


setup.py
```diff
@@ -28,18 +28,8 @@
...
-          'Programming Language :: Python :: 3.3',
-          'Programming Language :: Python :: 3.4',
-          'Programming Language :: Python :: 3.5',
-          'Programming Language :: Python :: 3.6',
-          'Programming Language :: Python :: 3.7',
-          'Programming Language :: Python :: 3.8',
-          'Programming Language :: Python :: 3.9',
-          'Programming Language :: Python :: 3.10',
-          'Programming Language :: Python :: 3.11',
+          'Programming Language :: Python :: 3.12',
```
akharit, line 32: Add back python 3.10 support to ensure azure cli can consume this

azure/datalake/store/lib.py
```diff
@@ -294,10 +143,18 @@ def __init__(self, store_name=default_store, token=None,
             __version__)
         self.req_timeout_s = req_timeout_s
 
+    def get_refreshed_bearer_token(self):
+        # Check if the token is about to expire in 300 seconds and refresh it if necessary
+        if self.AccessToken is None or time.time() > self.AccessToken.expires_on - 300:
+            self.AccessToken = self.token_credential.get_token(self.scopes)
+        return self.AccessToken.token
```
rayluo, line 150: FWIW, when using MSAL or its upper-level library Azure Identity, you can simply call their token acquisition method (`get_token()` in this case) unconditionally, and the library will take care of the refresh logic. I believe that is the same for Azure Identity. CC: @xiangyan99 

```suggestion
        self.AccessToken = self.token_credential.get_token(self.scopes)
        return self.AccessToken.token
```

akharit: My local test code using DefaultAzureCredential was eventually ending up on using AzurePowershellCredential which doesn't support 
In-memory token caching according to the documentation. It seems to depend on the kind of token. 

See table 
https://github.com/Azure/azure-sdk-for-python/blob/ded844291d88a9fb2d4877f332fcc84e7361b6d3/sdk/identity/azure-identity/TOKEN_CACHING.md#credentials-supporting-token-caching

rayluo: OK, then. I'll defer to @xiangyan99 :-)

dylanw-oss:
@akharit, thanks for fixing it! Do you have an ETA when a new package will be available?

azure/datalake/store/lib.py
```diff
@@ -269,22 +119,19 @@ class DatalakeRESTInterface:
         'REMOVEDEFAULTACL': ('put', set(), set())
     }
 
-    def __init__(self, store_name=default_store, token=None,
-                 url_suffix=default_adls_suffix, api_version='2018-09-01', req_timeout_s=60, **kwargs):
+    def __init__(self, store_name=default_store, token_credential=None,
```
xiangyan99, line 122: Is token credential the only credential type you support? e.g. do you support api key auth?

xiangyan99: Seems like api_version & req_timeout_s should be keyword only arguments.

and removing **kwargs is a breaking change.

Is this by design?

azure/datalake/store/lib.py
```diff
@@ -296,8 +143,10 @@ def __init__(self, store_name=default_store, token=None,
 
     @property
     def session(self):
+        bearer_token = self.token_credential.get_token(self.scopes).token
```
xiangyan99, line 146: As you noticed, token_credential could end up with Cli/Pws credential. We should cache it.

setup.py
```diff
@@ -48,13 +38,10 @@
                                       ]),
       install_requires=[
           'cffi',
-          'msal>=1.16.0,<2',  # http_cache was introduced in MSAL 1.16.0
           'requests>=2.20.0',
+          'azure-identity',
```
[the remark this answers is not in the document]
xiangyan99: The question is how user can use your library. Is TokenCredential required? If yes, you need 'azure-identity', if it is optional, then no, you don't need it in the requires.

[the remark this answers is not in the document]
akharit: The TokenCredential interface is required. So, leaving this in.

I was thinking of a customer coding a custom class with which satisfies the TokenCredential interface. Python type system allows it via duck typing, but if I follow that logic, any package will never have any dependency. 

I think I over complicated this because the previous code copied the interface of DataLakeCredential..

azure/datalake/store/lib.py
akharit: Breaking changes are fine, as we are making other breaking changes as well(like tokencredential and scope). 

I removed the kwargs as it was mainly forwarding the arguments earlier to the auth function. 

But see your point on making api_version and req_timeout_s keyword only args.

setup.py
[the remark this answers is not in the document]
xiangyan99: Good point. Yes. User can provide a custom credential that implements TokenCredential protocol. In this case, azure-identity library is not required.

Feel free to remove it from the install_requires.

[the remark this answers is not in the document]
akharit: Adding it in extras_require

azure/datalake/store/lib.py
akharit: Discussed with Xiang and we do want some caching here in case customer is using one of the methods without caching. and there is no harm in adding caching for other methods.

azure/datalake/store/lib.py
akharit: Added back the caching logic

beansrowning:
Ha, well this is silly but I had been using lib.auth() for most of my applications because it worked so seamlessly for interactive auth via device code without having to install an additional dependency in azure CLI

```py
from azure.datalake.store import lib
import polars as pl
token = lib.auth(
    tenant_id=os.environ["SP_TENANT_ID"],
    resource = 'https://storage.azure.com/'
)
# <interactive prompt for devicecode login>

# Ex. downstream, to read in data via polars
opts = {
    "tenant_id": token.token["tenant"],
    "client_id": token.token["client"],
    "token": token.token["access_token"],
    "account_name": "my_account"
}

lazy_data = pl.scan_parquet("az://some.path/here", storage_options=opts)
```

perusing the azure-identity documents and trying a couple dozen combinations, I finally figured out a replacement. 

```py
from azure.identity import DefaultAzureCredential

token_credential = DefaultAzureCredential(exclude_interactive_browser_credential=False, interactive_browser_tenant_id=os.environ["SP_TENANT_ID"])
access_token = token_credential.get_token('https://storage.azure.com/.default', tenant_id=os.environ["SP_TENANT_ID"])
```

Hopefully, this will save someone reading a couple of hours of headaches trying to get this working.

PR #335, opened 2025-04-09
To automate Python SDK Test Run

Ayushi-Gupta7:
---

### General Guidelines

### DIFF
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -16,6 +16,7 @@ RESOURCE_GROUP_NAME = fake_settings.RESOURCE_GROUP_NAME
 RECORD_MODE = os.environ.get('RECORD_MODE', 'all').lower()
 AZURE_ACL_TEST_APPID = os.environ.get('AZURE_ACL_TEST_APPID')
 CLIENT_ID = os.environ['azure_service_principal']
+ROOT_DIR = os.environ['ROOT_DIR']
 
 if RECORD_MODE == 'none':
     STORE_NAME = fake_settings.STORE_NAME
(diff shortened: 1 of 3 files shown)


Ayushi-Gupta7:
@microsoft-github-policy-service agree company="Microsoft"

Regards
Ayushi Gupta

tests/settings.py
```diff
@@ -16,6 +16,7 @@
 RECORD_MODE = os.environ.get('RECORD_MODE', 'all').lower()
 AZURE_ACL_TEST_APPID = os.environ.get('AZURE_ACL_TEST_APPID')
 CLIENT_ID = os.environ['azure_service_principal']
+ROOT_DIR = os.environ['ROOT_DIR']
```
Ayushi-Gupta7, line 19: @akharit Should we add in readme to send this new environment variable?

tests/test_multithread.py
```diff
@@ -168,7 +168,7 @@ def test_download_path(azure):
         down = ADLDownloader(
             azure,
             lpath="/lpath/test/testfolder",
-            rpath='/' + test_dir.name,
+            rpath='/'/test_dir,
```
Ayushi-Gupta7, line 171: This was tested from past few release and earlier code was causing issue. With new change, this wont fail

tests/settings.py
akharit: It will raise an exception as is so we don't need to, but it is a nicer UI. 

We can also have a default here so we have a random or constant root dir if this is not set.

akharit (approved):

tests/settings.py
Ayushi-Gupta7: Sure I can add a default value in case it is not provided

Ayushi-Gupta7: @akharit  do I need to change history.rst file?

akharit: Not needed for test only changes.

